PR #22, opened 2021-11-03
Allow ioki-objects to be used in array types

A5308Y:
instead of requiring a hash. Similar to type :object.


A5308Y:
@rmehner This is the fix for your problem with arrays.

rmehner (approved):

ans-ioki (approved):

PR #23, opened 2021-11-08
Less defaults for Ioki::Configuration.new

A5308Y:
These were the guiding principles for the refactoring. Based on previous discussions with @ans-ioki:

1. A new config should be as comfortable as possible to create in the console -> `Ioki::Configuration.from_env`
1. Each attribute of the config should be changeable by `client.config.attribute = setting`
1. config.reset! should read env-variables to allow changing during a console-development workflow
1. The VCR specs have to be able to read from env so they can send real requests.
1. The config should not know about multiple environments, because future developers will be expected to only work with one API.
1. Don't read from envs per default and be as explicit as possible to ease tracing of values to their origin.

The most important change for me is probably not reading from env per default.

### DIFF
--- a/lib/ioki/configuration.rb
+++ b/lib/ioki/configuration.rb
@@ -4,9 +4,13 @@ require 'ioki/std_out_logger'
 
 module Ioki
   class Configuration
-    DEFAULT_HTTP_ADAPTER            = 'faraday'
-    DEFAULT_API_BASE_URL            = 'https://app.io.ki/api/'
-    DEFAULT_API_VERSION             = '20210101'
+    DEFAULT_VALUES =
+      {
+        http_adapter: :faraday,
+        api_base_url: 'https://app.io.ki/api/',
+        api_version:  '20210101',
+        language:     'de'
+      }.freeze
 
     CONFIG_KEYS = [
       :http_adapter,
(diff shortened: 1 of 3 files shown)


ans-ioki (approved):
It streamlines and improves the config, so we can merge it as it is - still i think we can make it even more intuitive by elaborating on the client.

Instead of calling
```
my_client = Ioki::Client.new(Ioki::Configuration.from_env('DRIVER'), Ioki::DriverApi)
```

i'd like to only call

```
my_client = Ioki::Client.new(Ioki::DriverApi)
```

How about changing the "Parameter Position" - i think the "EndPoint Set Injection" feels way more important than the config. So the signature would be Ioki::Client#initialize(api, config = nil)

Or even better/additionally: Have a Ioki::[Platform/Passenger/Driver]Client.new convenience class, which inherits from Client and overrides the initializer by passing the defaults along.

```
my_client = Ioki::PlatformClient.new
```

My reasoning is: When i add the gem into a new Rails project and have setup the config values in env variables, i want nearly to no code to get a working client outside of an initializer. I think i never communicated this thought properly.
The whole point in having a "configurable configuration" was that clients will use it over again and again.
When you do not write a multi tenancy app you'd want to setup an initializer, assemble the "top level config" and forget about it.

A5308Y:
> Have a Ioki::[Platform/Passenger/Driver]Client.new convenience class, which inherits from Client and overrides the initializer by passing the defaults along.

I tried to avoid something like that, because of an earlier discussion we had about this, but I like this. I'll add the classes.

PR #26, opened 2021-11-08
Implement specific clients

A5308Y:
Based on https://github.com/ioki-mobility/ioki-ruby/pull/23#pullrequestreview-799933336


A5308Y:
I'm still hesitant to embed the assumption that devs will always store their config options in ENV. I think Rails credentials are generally a better option for example.
I do realise that this is how ioki sets up projects generally, but there might be exceptions, right? So if you  confirm this for me @ans-ioki I'll also implement the "Parameter Position" suggestion.

> you'd want to setup an initializer, assemble the "top level config" and forget about it.
I'd argue that's still something that's easily done by defining how to find the configuration params once.

One more question: Do we need the default values at all? I think it would be easier to understand how `initialise` and `reset!` work if we require a complete set of values in env or explicitly.

ans-ioki:
I think you did still not get the point i tried to explain:
The Ioki::Configuration class was once built to hold values "top level" (on an instance living in Ioki#config). No manual instantiation would be needed to hold the default set of credentials. How they get there is usually not relevant. An initializer could do this. It can do this from env. It can do this however it likes.

As the coder who sets up the Ioki::Client i MAY want to pass a handcrafted (with #new built) config. But i most of the times just don't want to do this (as i usually do not write a multitenancy app) and the one from Ioki#config gets cloned.

If i setup a new Rails App i will write an initializer that contains
```
Ioki::configure do |c|
  c.api_token = #... own code using rails secrets or env or whatever
end
```

But from then on i want to write

```
client = Ioki::Client.new(Ioki::DriverApi)
# or even
client = Ioki::DriverClient.new
```

ans-ioki (approved):

PR #29, opened 2021-11-15
Move http_adapter_class to config

A5308Y:
to simplify handling of parameters a bit.

The config has everything it needs to decide on the http adapter class so there's no reason to pass it on as a string or symbol IMO.


ans-ioki (approved):
most likely obsolete then :D

A5308Y:
Closed in favour of https://github.com/ioki-mobility/ioki-ruby/pull/38

PR #32, opened 2021-11-16
Refactor url_elements

A5308Y:
Also simplifies error handling by relying on ruby's existing error and
removing an unnecessary argument which was only used for an error
message. The information is still there for debugging because in the
backtrace, because we introduced more classes.

### DIFF
--- a/spec/ioki/apis/endpoints/endpoints_spec.rb
+++ b/spec/ioki/apis/endpoints/endpoints_spec.rb
@@ -3,33 +3,36 @@
 RSpec.describe Endpoints do
   describe '.url_elements' do
     it 'creates a concrete path from one with placeholders and corresponding string args' do
-      expect(described_class.url_elements('platform', [:id], 'RIDE_ID')).
-        to eq ['RIDE_ID']
+      expect(described_class.url_elements([:id], 'RIDE_ID')).to eq ['RIDE_ID']
     end
 
     it 'keeps path fragments that are not meant to be interpolated' do
-      expect(described_class.url_elements('platform', ['ride', :id], 'RIDE_ID')).
-        to eq %w[ride RIDE_ID]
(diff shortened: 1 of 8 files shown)


spec/ioki/apis/endpoints/endpoints_spec.rb
```diff
@@ -24,12 +23,24 @@
...
     it 'raises an error if the argument does not respond_to the interpolation method' do
       expect do
-        described_class.url_elements(
-          'platform',
-          [:id],
-          instance_double(Ioki::Model::Platform::RidePassenger, 'non_id_object')
-        )
-      end.to raise_error(ArgumentError, "#'platform': argument 1 must respond to :id or to be a string")
+        described_class.url_elements([:id], 1)
+      end.to raise_error(NoMethodError, /undefined method `id' for 1:Integer/)
```
A5308Y, line 27: I think the NoMethodError is as good as the custom error here and it allows to simplify the code.

spec/ioki/apis/endpoints/endpoints_spec.rb
```diff
@@ -3,33 +3,36 @@
...
     it 'raises an error if the argument does not respond_to the interpolation method' do
-      expect do
-        described_class.url_elements(
-          'platform',
-          [:id],
-          instance_double(Ioki::Model::Platform::RidePassenger, 'non_id_object')
-        )
-      end.to raise_error(ArgumentError, "#'platform': argument 1 must respond to :id or to be a string")
+      expect { described_class.url_elements([:id], 1) }.
+        to raise_error(NoMethodError, /undefined method `id' for 1:Integer/)
```
A5308Y, line 21: I think the NoMethodError is as good as the custom error here and it allows to simplify the code a bit.

ans-ioki (approved):

PR #38, opened 2021-11-18
Removes custom logger and HttpAdapter subclasses, uses Faraday directly

eikes:
Faraday provides some tools which were re-implemented in this gem: notably logging. The built-in logging ability of a Faraday::Connection is superior to a custom solution. Any custom logging code is removed and replaced with the Faraday logger middleware.

The Faraday::Connection is also able to have all headers preconfigured, so they don't need to be set manually in every request.

The request and response can be transparently en- and de-coded to and from JSON using the faraday JSON middleware. 

The custom test adapter was removed because it wasn't used anywhere. Faraday has a test stub adapter built in, so that is the preferred way of inspecting requests and providing custom responses.

As a user of this gem, I would like to be able to provide my own Faraday::Connection with the Test Adapter, so I can run tests against my own responses. These changes make it possible to do so.


A5308Y (commented):
I like this a lot! 
I think @ans-ioki had specific ideas for logging and customising adapters. Is this compatible with your ideas @ans-ioki?

eikes:
I realize this is a big PR. I really got into it, learned a lot about the gem while I was at it. 

I think it would be beneficial if we would set up a meeting-call where we could go through the changes. Otherwise I'd more than happy to answer any questions here in the PR.

ans-ioki:
All well! Be bold and move the code. Regarding the decoupled http adapter: Back then, when i initially set up the stuff, i planned to not use VCR and aimed on having a dedicated testadapter, like stripe, which to some extend is able to carry internal state over multiple requests. But go ahead and evolve it this, i like the changes. Andreas H, as the main contributor, should be able to approve incoming PRs mostly on his own.

A5308Y (approved):

A5308Y:
> I think it would be beneficial if we would set up a meeting-call

I think I understand the changes @eikes and I would just merge this now that @ans-ioki gave some feedback as well. Is there something you would like to discuss specifically?

eikes:
Looks good to me! Thank you for letting me contribute this large PR.

PR #43, opened 2021-11-24
Add missing platform models and attributes

A5308Y:
# Todos

- [x] Support all GET-paths of the platform API so schaufel can use all the new models for the dashboard.
- [x] Transform `unspecified_models` in `openapi_spec.rb` to `unvalidated` on the model.
- [x] Try to avoid calling `Ioki::Model::Platform` all the time
- [x] Reenable all model checks and find a way to only check for validity and not completeness (passenger, driver) for now. It's probably best to fix all existing models and not add the other models for now so we have missing models but not partially supported models. That way it's more obvious if something is not supported yet.
- [x] Read `specification_scope` and `schema_path` from the model instead of passing it on.
- [x] Find a way not to have to write: `def self.specification_scope; nil; end`
- [x] Require all models in ioki-model
- [x] Explain/document `unvalidated` and maybe find a better name.
- [x] Check what happens for dependent "object"-attributes when the model_class is omitted. -> It'll be a Hash.
- [x] Manually check that all unvalidated models inherit from the correct base class.
- [x] Ensure all date_time objects are date_time
- [x] Rubocop

Edit:
This is meant as a quick fix so @rmehner and @eikes don't have to wait even longer. See https://github.com/ioki-mobility/ioki-ruby/issues/52 for a follow-up issue.


rmehner (approved):
Looks good. Would be nice to get that in a release soon, so I can consume it to make the form dynamic.

Going forward I'd suggest to add smaller PRs on top of this (says the person who previously opened up a massive PR ...)

PR #44, opened 2021-11-25
Decode JSON before logging, so no Encoding::UndefinedConversionError is raised when logging Umlaute

eikes:


rmehner (approved):

rmehner:
Ha, just ran into this in another project as well. Fun times.

PR #45, opened 2021-11-25
Fixes log filter to actually work

eikes:
The filter regexp didn't match the actual headers. Now there is a test which verifies that...

### DIFF
--- a/lib/ioki/http_adapter.rb
+++ b/lib/ioki/http_adapter.rb
@@ -10,20 +10,22 @@ module Ioki
       ::Faraday.new(config.api_base_url, headers: headers(config)) do |f|
         f.adapter :net_http
 
-        # The order of execution of the middleware is FiFo:
-        # 1) parsed as JSON (and implicitly is encoded properly as UTF-8)
-        # 2) logging occurs
+        f.request :authorization, 'Bearer', -> { config.api_token }
+
+        # The placement of the middleware here determines the execution order
+        # It's important, that the logger is defined before the :json middleware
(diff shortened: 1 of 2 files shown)


lib/ioki/http_adapter.rb
```diff
@@ -10,20 +10,20 @@ def self.get(config)
       ::Faraday.new(config.api_base_url, headers: headers(config)) do |f|
         f.adapter :net_http
 
+        f.request :authorization, 'Bearer', -> { config.api_token }
+        f.request :json # encode req bodies as JSON
```
rmehner, line 14: Why is this back on top again? Doesn't this revert the changes of #44?

eikes: These two are the middleware for the outgoing `request`. The order of the middleware for the incoming `response` is what I changed in #44. The thing is, that the logger inspects both `resquest` and `response` even if it is only called on the response (don't ask me why) and I noticed, that the authorization header would not be logged if it is configured in this setup after the logger. I was not able to test yet how this order affects the logging of the outgoing request bodies, such as a POST request body.

I can take a look after the weekend though.

eikes:
@rmehner I checked and the request body for POST / PATCH requests was not parsed before logging. Thanks for pointing that out. It did work regardless, even for non-ascii characters already though, because the request was already using UTF-8. Anyhow, I changed it and the request formatting of the json body in the log now matches the response.

A5308Y (approved):

PR #48, opened 2021-11-29
Don't require version on ride creation

rmehner:
Also removes accidental double definition of type on passenger ride.


lib/ioki/model/passenger/ride.rb
```diff
@@ -54,7 +53,7 @@ class Ride < Base
         attribute :vehicle, type: :object, model_class: Ioki::Model::Passenger::Vehicle, on: [:read, :create, :update]
         attribute :vehicle_reached_dropoff, type: :boolean, on: [:read, :create, :update]
         attribute :vehicle_reached_pickup, type: :boolean, on: [:read, :create, :update]
-        attribute :version, type: :integer, on: [:read, :create, :update]
+        attribute :version, type: :integer, on: [:read, :create, :update], omit_if_blank_on: [:create]
```
ans-ioki, line 56: imho just
```
attribute :version, type: :integer, on: [:read, :update]
```

A5308Y (approved):

PR #49, opened 2021-11-30
Version bump to 0.6.2

A5308Y:


rmehner:
Is the changelog even needed, given that we have that in releases? https://github.com/ioki-mobility/ioki-ruby/releases

rmehner (approved):

PR #50, opened 2021-11-30
Fixes endless autopagination for empty results with last_page: false

eikes:
When there are no results, the API returns last_page: false for the page 1 auto_paginate should not continue to request data when there are no results

https://github.com/kaminari/kaminari/issues/962
https://github.com/kaminari/kaminari/issues/582


eikes:
@ans-ioki, could you take a look at this PR and maybe merge it? Thank you!

eikes:
This has been resolved on the API side

PR #54, opened 2021-12-17
Runtime attribute class resolution

eikes:

PR #55, opened 2022-01-06
Fix version being required on create for rides

rmehner:
This was accidentally introduced in #54 and prevented us from creating
rides, since they don't have a version then as they are just about to be
created.

In the long run, we might wanna sort those attributes alphabetically, so that these kind of double definitions are easier to spot.


eikes:
What does the version field represent?

rmehner:
The current version of the ride. It will be increased if any of the attributes change. This is important for conflict detection as every update needs to provide the version it tries to update or else it will fail.

Also note that version had been defined already a couple of lines down and was accidentally defined again there, but with `create` scope added.

eikes (approved):
Looks good to me!

PR #57, opened 2022-01-11
Enables iterating over index endpoint results in a given block

eikes:
In stead of creating a very large array I would prefer to iterate over the results directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Basic usage of a `Ioki::Client`
   products = platform_client.products
   # returns a list of Ioki::Model::Platform::Product instances
 
-  stations = platform_client.stations(products.first, options: { auto_paginate: true })
+  stations = platform_client.stations(products.first, auto_paginate: true)
   # stations are a scoped endpoint within products, to the interface requires
   # either a product or a product id as the first parameter.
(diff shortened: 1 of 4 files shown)


eikes:
This allows us to do the following:

```ruby
client = Ioki::PlatformClient.new
product = client.products.first
stations = client.stations(product, options: { auto_paginate: true }) do |station|
  puts station.id
end
```

A5308Y:
LGTM. And sorry for the delay. I'll probably not be working on ioki-ruby this month.

A5308Y (approved):

rmehner:
Sorry, didn't have the time earlier. Since we're not streaming, I don't really see the advantage and it loses the clearness of "each" in the call. But that might be just me. I'm not opposed to it, just wondering why this is better than good old ".each do", which signals iteration.

eikes:
@rmehner The advantage is, that the `results` array is not created in its entirety, because it could become rather large. Instead each page is requested one by one and the results are yielded. So in a sense, this is a sort of streaming, but rather on a index page basis.

rmehner:
> Instead each page is requested one by one and the results are yielded. 

I read that part of the code wrong. That makes a lot of sense indeed. Thanks for clarifying!

PR #59, opened 2022-01-20
Allow `present?` and `blank?` to be used in Rails

phylor:
`Ioki::Model::Base#blank?` has precedence over `Object#blank?` defined in Rails. Therefore it is currently not possible to use `present?` or `blank?` on any objects returned from the API within Rails. Rails defines `present?` as `!blank?`.

This is a bit suprising for developers familiar with Rails. Instead of writing `if ride.pickup.present?`, we have to resort to `unless ride.pickup.nil?`.

I'd therefore propose to rename `Ioki::Model::Base#blank?` to `Ioki::Model::Base#empty?`, or another name you're comfortable with.

### DIFF
--- a/lib/ioki/model/base.rb
+++ b/lib/ioki/model/base.rb
@@ -141,7 +141,7 @@ module Ioki
 
           next if definition.key?(:omit_if_blank_on) &&
                   Array(definition[:omit_if_blank_on]).include?(usecase) &&
-                  blank?(value)
+                  empty?(value)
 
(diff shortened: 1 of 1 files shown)


rmehner:
👍🏿  from me. Any thoughts on this @eikes?

A5308Y (approved):
I thought about a name a bit because `empty?` is defined for "collections" in Rails as well, but I think it's Ok to define it on a "single object" like this. So 👍

PR #61, opened 2022-01-26
Silence warning in specs

A5308Y:


A5308Y:
`remove_const` is a private method so `public_send` is not an option. This [thread](https://stackoverflow.com/questions/9236264/how-to-disable-warning-for-redefining-a-constant-when-loading-a-file) suggests 2 other ways of disabling the warning for our case, but I found `send` in specs acceptable and a little less work to understand.

eikes (approved):
LGTM!

PR #66, opened 2022-02-01
Add announcements to platform API

rmehner:


lib/ioki/apis/platform_api.rb
```diff
@@ -167,6 +167,16 @@ class PlatformApi
         actions:     { 'lock' => :patch, 'unlock' => :patch, 'set_pin' => :patch },
         path:        [API_BASE_PATH, 'products', :id, 'drivers', :id],
         model_class: Ioki::Model::Platform::Driver
+      ),
+      Endpoints::Index.new(
```
A5308Y, line 171: We have the `crud_enpoints` "macro" for stuff like this:
```ruby
Endpoints.crud_endpoints(
  :announcement,
  except: [:create, :update, :delete],
  model_class: Ioki::Model::Platform::Announcement
)
```

A5308Y (approved):
LGTM! Only thing is using the "macro".

lib/ioki/apis/platform_api.rb
rmehner: Thanks, done :)

PR #71, opened 2022-02-10
Exposes latest request in Endpoints::Index when block is passed for iteration

eikes:
When passing a block to the index endpoint I would like to be able to inspect the current request. This change allows that:

```ruby
platform_client.products do |product, response|
  puts product.id
  puts response.env.url
end
```

It is still possible to use the block without the response:

```ruby
platform_client.products do |product|
  puts product.id
end
```


rmehner:
Thank you. I think we should trigger a new release soon.

Issue #73, opened 2022-02-14
Prepare update of TaskList API


eikes:
There are some changes planned for the TaskList API, we need to make sure, that ioki-ruby will still work afterwards:

Those changes are planned:

1. calculated_point
    * nested "station" is now deprecated
    * attribute "station_id" was added
2. task_list_serializer
    * nested "matching_configuration" is now deprecated
    * attribute "matching_configuration_id" was added
3. task_serializer
    * nested "driver" is now deprecated. please note: the nested vehicle stays in place and the drivers id is accessible from the nested 
    * nested "pause" is now deprectated
    * attribute "pause_id" was added
    * nested "ride" is now deprecated
    * attribute "ride_id" was added

The deprecated models will get rendered as "null" after the next rollout and the whole key will get removed 8 weeks after.

eikes:
The main change is here:

https://github.com/ioki-mobility/ioki-ruby/pull/74/commits/544ecee9a5c4c143353e8356546d3037c8f6b435

eikes:
Regarding the 'task_serializer: nested "driver" is now deprecated': I don't see a driver in the Task, neither in the platform model nor in the OpenAPI specification.

PR #74, opened 2022-02-16
Platform api task list changes

eikes:
Make the changes required to address #73


eikes:
This PR will remain a draft until the changes have been made on the API side. Feel free to give me feedback beforehand.

ans-ioki (approved):

ans-ioki:
I think we should not wait until the stuff is gone. Let's make this the new version.

lib/ioki/model/platform/calculated_point.rb
```diff
@@ -22,7 +22,6 @@ class CalculatedPoint < Base
         attribute :waypoint_type,        on: :read, type: :string
         attribute :walking_duration,     on: :read, type: :integer
         attribute :walking_track,        on: :read, type: :string
-        attribute :station,              on: :read, type: :object, class_name: 'Station'
```
rmehner, line 25: The Platform API docs say that we do have a station in there (https://man.pages.ioki.com/api_docs/refs/master/platform-api.html#tag/Ride/paths/~1api~1platform~1products~1{product_id}~1rides/get). Is there a mismatch between docs and reality?

spec/ioki/driver_api_spec.rb
```diff
@@ -34,7 +34,7 @@
       end
 
       expect(driver_client.create_request_token(request_token, options)).
-        to eq(Ioki::Model::Driver::DriverRequestToken.new)
+        to be_a(Ioki::Model::Driver::DriverRequestToken)
```
rmehner, line 37: Nice little improvement :)

lib/ioki/model/platform/task.rb
```diff
@@ -15,9 +15,9 @@ class Task < Base
         attribute :completer, on: :read, type: :object
         attribute :completer_id, on: :read, type: :string
         attribute :completer_type, on: :read, type: :string
-        attribute :pause, on: :read, type: :object, class_name: 'Pause'
+        attribute :pause_id, on: :read, type: :string
         attribute :rejectable, on: :read, type: :boolean
-        attribute :ride, on: :read, type: :object, class_name: 'Ride'
+        attribute :ride_id, on: :read, type: :string
```
rmehner, line 20: Is this no longer nested? Because with other cases it usually is.

rmehner (approved):
Looks good for me, some questions remain, but those are minor.

lib/ioki/model/platform/calculated_point.rb
rmehner: Alright, good to know, since we rely on some of those nested things in Schalter, so we'd need to double check (at least for Passenger API)

lib/ioki/model/platform/task.rb
eikes: Same as above :D

spec/ioki/driver_api_spec.rb
eikes: Thank you, I find it more readable...

lib/ioki/model/platform/calculated_point.rb
eikes: Yes, @rmehner, you should check if you rely on the nested model. 

@ans-ioki When will those API changes go live? Does it not make sense to wait until then?

rmehner: We do for Passenger API (booking & station at least), @ans-ioki does this mean we need to adapt this for us as well?

rmehner:
:wave: this looks pretty stale by now. Have the API changes been made?

lib/ioki/model/platform/calculated_point.rb
eis-ioki: There is only the station_id now.

eis-ioki:
The changes have been made. I checked.

eis-ioki:
@rmehner And finally I was able to address this. Sorry for the delay!

rmehner (approved):

rmehner:
Thanks :)

PR #75, opened 2022-02-22
Adds :provider_id to Ioki::Model::Platform::User model 

eikes:
The Platform API has thankfully added the `:provider_id` attribute to the `platform_api--v20210101--user` definition. This PR adds it to the Ioki::Model::Platform::User model.

Additionally this PR adds unit tests for this model, for which a custom matcher has been implemented.

### DIFF
--- a/lib/ioki/model/platform/user.rb
+++ b/lib/ioki/model/platform/user.rb
@@ -4,24 +4,24 @@ module Ioki
   module Model
     module Platform
       class User < Base
-        attribute :email,             type: :object,  on: [:read, :create, :update],
-omit_if_blank_on: [:create, :update], class_name: 'UserEmail'
-        attribute :external_id,       type: :string,  on: [:read, :create, :update],
-omit_if_blank_on: [:create, :update]
-        attribute :first_name,        type: :string,  on: [:read, :create, :update],
-omit_if_blank_on: [:create, :update]
-        attribute :last_name,         type: :string,  on: [:read, :create, :update],
-omit_if_blank_on: [:create, :update]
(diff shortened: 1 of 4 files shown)


lib/ioki/model/platform/user.rb
```diff
@@ -4,24 +4,24 @@ module Ioki
...
+                  class_name:       'UserEmail'
+        attribute :external_id, type: :string, on: [:read, :create, :update], omit_if_blank_on: [:create, :update]
+        attribute :first_name, type: :string, on: [:read, :create, :update], omit_if_blank_on: [:create, :update]
+        attribute :last_name, type: :string, on: [:read, :create, :update], omit_if_blank_on: [:create, :update]
+        attribute :locale, type: :string, on: :read
+        attribute :lock_reason, type: :string, on: :read
+        attribute :lock_type, type: :string, on: :read
+        attribute :locked_at, type: :date_time, on: :read
+        attribute :phone_number, type: :string, on: [:read, :create, :update], omit_if_blank_on: [:create, :update]
+        attribute :provider_id, type: :string, on: :read
```
eikes, line 20: This is the new line. All others have simply been reformatted.

rmehner (approved):

rmehner:
Nice with the matcher. I'm usually torn with these kind of tests, since they repeat the implementation, but for a library I guess it makes sense and it reads nice with those matchers.

eikes:
I agree. But on the other hand I don't want to commit a change that is untested. Maybe using these matchers we can cover the other models as well?

PR #77, opened 2022-02-28
Phone verification endpoints

phylor:
Add configuration to allow to call phone verification endpoints.

### DIFF
--- a/lib/ioki/model/platform/phone_verification.rb
+++ b/lib/ioki/model/platform/phone_verification.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Platform
+      class PhoneVerification < Base
+        attribute :phone_number, on: :create, type: :string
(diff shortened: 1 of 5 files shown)


lib/ioki/model/platform/phone_verification.rb
```diff
@@ -0,0 +1,12 @@
...
+module Ioki
+  module Model
+    module Platform
+      class PhoneVerification < Base
+        attribute :phone_number, on: :create, type: :string
+        attribute :code, on: :create, type: :string
+      end
+    end
+  end
+end
```
rmehner, line 12: Have you seen #75? This introduces matchers to have at least some spec coverage for this. What do you think of this?

rmehner (approved):

lib/ioki/model/platform/phone_verification.rb
phylor: Added specs as suggested!

PR #80, opened 2022-03-08
Allow email to be provided when creating user

phylor:
Allows email-only users to be created. For that we need to be able to provide the email address `on: :create`.

### DIFF
--- a/lib/ioki/model/platform/user_email.rb
+++ b/lib/ioki/model/platform/user_email.rb
@@ -10,9 +10,9 @@ module Ioki
         unvalidated true
 
         attribute :confirmed,     on: :read, type: :boolean
-        attribute :email_address, on: [:read, :update], type: :string
-        attribute :newsletter,    on: [:read, :update], type: :boolean
-        attribute :receipt,       on: [:read, :update], type: :boolean
+        attribute :email_address, on: [:create, :read, :update], type: :string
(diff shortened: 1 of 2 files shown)


lib/ioki/model/platform/user_email.rb
```diff
@@ -10,9 +10,9 @@ class UserEmail < Ioki::Model::Base
         unvalidated true
 
         attribute :confirmed,     on: :read, type: :boolean
-        attribute :email_address, on: [:read, :update], type: :string
-        attribute :newsletter,    on: [:read, :update], type: :boolean
-        attribute :receipt,       on: [:read, :update], type: :boolean
+        attribute :email_address, on: [:create, :read, :update], type: :string
+        attribute :newsletter,    on: [:create, :read, :update], type: :boolean
+        attribute :receipt,       on: [:create, :read, :update], type: :boolean
```
rmehner, line 15: IIRC this makes these attributes required on create. Have you checked that?

phylor: Yes and no. It means that they are serialized no matter what. It is still fine however to send `nil` as a value, even if they are of type `boolean`. It is also fine not to specify them when creating the `UserEmail` model, i.e. `Ioki::Model::Platform::UserEmail.new(email: 'myemail@example.com')` is fine.

PR #81, opened 2022-03-16
Refine Vehicle model for Operator API

tom-ioki:
With this PR vehicle attributes will also be available on `create` and `update`


rmehner (approved):
Also improved readability by introducing sensible formatting. Well done! 👏🏿

PR #82, opened 2022-03-21
Add more webhooks support

eikes:
Webhooks require a validation of the `x-signature` HTTP header which is implemented here.

Additionally a helper class to convert the incoming webhook event into a Ioki::Webhooks::Model model is provided.


lib/ioki/webhooks/event.rb
```diff
@@ -0,0 +1,39 @@
...
+        camelize data['type']
+      end
+
+      def action
+        event_type&.split('.')&.last
+      end
+
+      private
+
+      def camelize(term)
```
A5308Y, line 27: I'm wondering if it would make sense to copy the Rails' implementation as that's what I'd expect most people using the gem are familiar with. Otherwise there might be subtle differences.

See https://apidock.com/rails/v6.1.3.1/String/camelize. Rails 7's is harder to read in my opinion: https://github.com/rails/rails/blob/de53ba56cab69fb9707785a397a59ac4aaee9d6f/activesupport/lib/active_support/inflector/methods.rb#L96.

Maybe it would also be nice to extract the function to a module as `camelize` is a concept that makes sense for all parts of the app.

What do you think @eikes?

README.md
```diff
@@ -114,6 +114,41 @@ the values in `Ioki.config`. Ioki.config will respect the values set up by
...
+
+Webhooks are a way to receive updates about changes when they happen on the
+backend servers. They are **pushed** to your application, where the other APIs
+are **pulled** from the backend. In this regard they are very different from
+the Platform/Driver/Passenger/Operator API. What they have in common, is that
+the data is serialized in a specific manner, which ioki-ruby can turn into a
+model in the Ioki::Model::Webhooks module.
+
+When you have the requests POST body in a `params` hash, you can convert that
+to a `Event` instance, which has a `model` attribute which is the deserialized
```
A5308Y, line 127:
```suggestion
to an `Event` instance, which has a `model` attribute which is the deserialized
```

A5308Y (approved):
LGTM I especially appreciate the addition to the Readme.

I just had a thought about `camelize`, but it's fine for me as well if you keep the current version.

lib/ioki/webhooks/event.rb
eikes: It is in fact copied from rails 7 here: https://github.com/rails/rails/blob/de53ba56cab69fb9707785a397a59ac4aaee9d6f/activesupport/lib/active_support/inflector/methods.rb#L69

But I had to remove everything related to inflections. Finally I had to replace the $1 and $2, because rubocop didn't like that.

In general though, I'd prefer to use ActiveSupport, but I was told not to. What are your thoughts on using integrating ActiveSupport? It would let us remove some reimplementations like this and everyone uses it anyways...

A5308Y: Huh. I looked at that method as well and didn't recognise it. Odd. I'd personally be Ok with ActiveSupport, but was also told not to use it and I'd stick with that. I think the next best option would be to have a module for "the stuff copied from Rails" with some comments and links to the original source.

eikes: OK, let's create a Ioki::Support module after merging this!

PR #84, opened 2022-03-22
Replaces 'pry' gem with state of the art 'debug' gem

eikes:


eikes:
The `debug` gem is becoming the default gem for debugging. In my humble opinion, it is fantastic. We should use it:

https://github.com/ruby/debug

A5308Y (approved):

PR #88, opened 2022-04-04
Expand support for Driver API

eikes:


lib/ioki/http_adapter.rb
```diff
@@ -21,7 +21,7 @@ def self.get(config)
           f.response :logger, config.logger, config.logger_options do |logger|
             logger.filter(/(X-Client-Secret: ")([^"]+)/, '\1[REMOVED]')
             logger.filter(/(Authorization: ")([^"]+)/, '\1[REMOVED]')
-            logger.filter(/("token":")([^"]+)/, '\1[REMOVED]')
```
A5308Y, line 24: I'm wondering if we should have both versions. AFAIK => is usually used for backwards compatibility in debug/log output. This might change at some point.

lib/ioki/model/driver/base.rb
```diff
@@ -7,11 +7,6 @@ class Base < Ioki::Model::Base
         def self.specification_scope
           'driver_api'
         end
-
-        attribute :id,         on: :read, type: :string
```
A5308Y, line 11: Not sure about this change. I usually like explicit over implicit, but after this change `Driver::Base` deviates from the rest of the base models. The "note that this does not inherit from" comment is not the best solution either I admit. Maybe we can find something else?

lib/ioki/apis/endpoints/endpoint.rb
```diff
@@ -4,16 +4,17 @@ module Endpoints
   class Endpoint
     attr_reader :resource, :action, :request_method, :path, :model_class
 
-    def initialize(resource, action:, request_method:, path:, model_class: nil)
+    def initialize(resource, action:, request_method:, path:, model_class: nil, method_name: nil)
```
A5308Y, line 7: 👍

A5308Y (approved):
Looks good to me. I'd appreciate some thoughts on the base model question though.

lib/ioki/model/driver/base.rb
eikes: IMHO inheritance isn't used correctly when some classes don't inherit from the class they should be inheriting from, semantically speaking. This goes for the other APIs as well. If not all classes share these attributes, then we could introduce another abstract class or I'd be happy to simply add the missing :id, :type and date attributes to all classes.

lib/ioki/http_adapter.rb
eikes: good point, it's easy enough to have both

lib/ioki/model/driver/base.rb
A5308Y: > IMHO inheritance isn't used correctly when some classes don't inherit from the class they should be inheriting from, semantically speaking

I agree.

I'm Ok with both versions (abstract class or adding fields), whatever is easier for you. But I think consistency is important, so I think it would be good to change the other models in this PR as well.

eikes: Cool! How about we merge this and I'll do that in a different PR?

A5308Y: :-) I just tried to suggested to it in **this** PR, because I'd want to avoid a state where inheritance is used differently in the different "APIs". But hey, I'm surprised I'm still being asked anyway :-). So. Sure. Go ahead :-)

PR #89, opened 2022-04-05
Operator Api: add show endpoint for product

tom-ioki:
This PR introduces the `show` endpoint for the Operator API.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -7,15 +7,11 @@ module Ioki
   class OperatorApi
     API_BASE_PATH = 'operator'
     ENDPOINTS = [
-      Endpoints::Index.new(
-        :products,
-        base_path:   [API_BASE_PATH],
-        model_class: Ioki::Model::Operator::Product
-      ),
-      Endpoints::Show.new(
+      Endpoints.crud_endpoints(
         :product,
         base_path:   [API_BASE_PATH],
-        model_class: Ioki::Model::Operator::Product
+        model_class: Ioki::Model::Operator::Product,
+        except:      [:create, :update, :delete]
       ),
       Endpoints.crud_endpoints(
         :vehicle,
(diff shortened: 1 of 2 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -12,6 +12,11 @@ class OperatorApi
         base_path:   [API_BASE_PATH],
         model_class: Ioki::Model::Operator::Product
       ),
+      Endpoints::Show.new(
+        :product,
+        base_path:   [API_BASE_PATH],
+        model_class: Ioki::Model::Operator::Product
+      ),
```
eikes, line 19: Looks good, you could also do this:
```ruby
      Endpoints.crud_endpoints(
        :product,
        base_path:   [API_BASE_PATH],
        model_class: Ioki::Model::Operator::Product,
        except:      [:create, :update, :delete]
      ),
```

tom-ioki:
@eikes could you have another look at this please 🙏  Thx

eikes:
Looks good to me!

eikes:
For extra credit: Add some tests!

tom-ioki:
Sorry for the late response.

Not sure if it makes that much sense to have specs for this change. The `OperatorApi` class contains two constants and that's it. Also testing for the content of the `ENDPOINTS` constant is rather tedious and impractical.

Other than that, there are no specs for `DriverApi`, `PassengerApi` and `PlatformApi` 🤷 

Is it fine, if I skip the specs for this?

eikes:
I was referring to tests like these:

https://github.com/ioki-mobility/ioki-ruby/blob/main/spec/ioki/platform_api_spec.rb#L51-L73

eikes:
I see that the Operator API has no tests yet, so it's a good time as any to create that file. If you don't find the time, we should at least create an issue for that.

tom-ioki:
Ah I see! That's where the specs are :D I was looking for them in `/spec/ioki/apis/`
Alright, I will add some specs then!

eikes:
Looks superbe to me! Thanks for adding that spec file!

tom-ioki:
Awesome! Thank you for the review! Can I merge it then?

eikes:
Yes!

PR #90, opened 2022-04-15
Platform API model add places endpoints and improvements for :create and :update serialization

eikes:


A5308Y (approved):
I'd really have to check the API-documentation in detail to review this PR properly, apart from that the changes look good though. So 👍

PR #91, opened 2022-04-15
Adds current_ride endpoint to passenger API

eikes:


spec/ioki/passenger_api_spec.rb
```diff
@@ -31,8 +31,7 @@
         result_with_index_data
       end
 
-      expect(passenger_client.products(options)).
-        to eq([Ioki::Model::Passenger::Product.new])
+      expect(passenger_client.products(options)).to all(be_a(Ioki::Model::Passenger::Product))
```
A5308Y, line 34: Looks nicer to me, but technically we lose some specificity because we don't test anymore that the result is the "empty" version of a product.

A5308Y (approved):

spec/ioki/passenger_api_spec.rb
eikes: True! But given that there are no attributes, this isn't very meaningful.

PR #93, opened 2022-04-21
All model class in an API module inherit from the same Base class

eikes:


eikes:
@A5308Y This is the change where the inheritance is fixed.

A5308Y (approved):

A5308Y:
Thanks!

PR #94, opened 2022-05-10
Add `User#notification_settings`

phylor:


rmehner:
> * When updating the notification settings, we also need to send the `user#version` attribute. I therefore created a new model `NotificationSettingsWithVersion`. The response includes the updated settings without the version. Not sure if this is the correct pattern to use or if there is another way to achieve the same?

This sounds like we're missing some context on why that version is not on the settings object itself, instead of the being the user version. There might be a good reason for it, maybe @ans-ioki can clear that up. I'd expect it to be a separate version, just like the rides, but I don't know all the internals.

Given the current state, I think it's the correct way to do this within this gem 👍🏿

rmehner (approved):

PR #96, opened 2022-05-19
Add line and line stop to plattform api

ozb-ioki:


lib/ioki/apis/platform_api.rb
```diff
@@ -70,6 +70,17 @@ class PlatformApi
...
         model_class: Ioki::Model::Platform::Pause
       ),
+      Endpoints.crud_endpoints(
+        :line,
+        base_path:   [API_BASE_PATH, 'products', :id],
+        model_class: Ioki::Model::Platform::Line
+      ),
+      Endpoints.crud_endpoints(
+        :line_stop,
+        base_path:   [API_BASE_PATH, 'lines', :id],
```
tom-ioki, line 80: I guess `line_stops` are also part of the product resource. In that case you need to adjust the `base_path` to:
```ruby
base_path: [API_BASE_PATH, 'products', :id, 'lines', :id]
```

tom-ioki (changes requested):
Great job so far, almost there.

Besides my comments, please have a look at the `spec/ioki/platform_api_spec.rb`. There you can add some specs for the added endpoints 👍

ozb-ioki:
> Great job so far, almost there.
> 
> Besides my comments, please have a look at the `spec/ioki/platform_api_spec.rb`. There you can add some specs for the added endpoints 👍 

Thank you for the review Tom. I will look on it and apply the changes.

lib/ioki/apis/platform_api.rb
ozb-ioki: Right, i had to push nested line_stops before, but changed it in the api after and did not apply here. Thanks.

tom-ioki (approved):
Nice! Looks good to me 🆗

PR #98, opened 2022-05-23
Add driver model and driver endpoints for Operator Api

tom-ioki:
This PR introduces the driver endpoints and refines the driver model for the Operator Api.

The driver endpoints for the operator api in triebwerk are still in review. But as soon as that PR is reviewed and merged, everything will be in place for this `ioki-ruby` PR.

### DIFF
--- a/spec/ioki/operator_api_spec.rb
+++ b/spec/ioki/operator_api_spec.rb
@@ -78,6 +78,7 @@ RSpec.describe Ioki::OperatorApi do
     it 'calls request on the client with expected params' do
       expect(operator_client).to receive(:request) do |params|
         expect(params[:url].to_s).to eq('operator/products/0815/vehicles')
+        expect(params[:method]).to eq(:post)
         [result_with_data, full_response]
       end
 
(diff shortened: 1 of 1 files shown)


spec/ioki/operator_api_spec.rb
```diff
@@ -111,4 +111,116 @@
...
+    it 'calls request on the client with expected params' do
+      expect(operator_client).to receive(:request) do |params|
+        expect(params[:url].to_s).to eq('operator/products/0815/drivers/4711')
+        result_with_data
+      end
+
+      expect(operator_client.delete_driver('0815', '4711', options)).
+        to eq(Ioki::Model::Operator::Driver.new)
+    end
+  end
```
ferdicus, line 177: hi there, sorry, but shouldn't this check for method as well, like the one below does for `lock`?
something like `expect(params[:method]).to eq(:delete)`?

ferdicus (approved):
thanks, left some comments

spec/ioki/operator_api_spec.rb
tom-ioki: valid point. I've added the missing specs (also for the vehicle endpoints). Thanks 🙏

PR #102, opened 2022-06-08
Add support for returning pagination info

phylor:
These changes allow `index` endpoints to return pagination info to the caller. Until now, an array of the first page is returned. Although it is already possible to request further pages, there is no way to know how many pages there are.

This allows the following (`paginate: true` as a new option):

```ruby
passenger_client.rides(params: {filter: "active", page: 2}, paginate: true)
```

Which would return (instead of an array):

```ruby
{
  data: [..rides..],
  meta: { page: 2, last_page: false }
}
```

When `paginate` is `false` (the default), an array is returned. That is, there's no breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,18 +35,22 @@ Basic usage of a `Ioki::Client`
   products = platform_client.products
   # returns a list of Ioki::Model::Platform::Product instances
 
-  stations = platform_client.stations(products.first, auto_paginate: true)
-  # stations are a scoped endpoint within products, to the interface requires
+  stations = platform_client.stations(products.first, paginate: true)
+  # stations are a scoped endpoint within products, so the interface requires
   # either a product or a product id as the first parameter.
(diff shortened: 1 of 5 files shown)


rmehner:
This all looks good, however I'd expect some tests over at https://github.com/ioki-mobility/ioki-ruby/blob/main/spec/ioki/apis/endpoints/index_spec.rb, which already tests autopagination and yielding.

phylor:
> This all looks good, however I'd expect some tests over at https://github.com/ioki-mobility/ioki-ruby/blob/main/spec/ioki/apis/endpoints/index_spec.rb, which already tests autopagination and yielding.

Sure, added specs! :slightly_smiling_face:

rmehner:
We should probably release a new version, to signal that we've [included quite a bit since the last release](https://github.com/ioki-mobility/ioki-ruby/compare/0.8.0...main)

PR #106, opened 2022-06-27
Logpay | Create payment methods

phylor:
Adds support to create a logpay payment method.

### DIFF
--- a/lib/ioki/apis/passenger_api.rb
+++ b/lib/ioki/apis/passenger_api.rb
@@ -132,6 +132,11 @@ module Ioki
         base_path:   [API_BASE_PATH],
         model_class: Ioki::Model::Passenger::PaymentMethod
       ),
+      Endpoints::Delete.new(
+        :payment_method,
+        base_path:   [API_BASE_PATH],
+        model_class: Ioki::Model::Passenger::PaymentMethod
+      ),
       Endpoints::Index.new(
         :service_credits,
         base_path:   [API_BASE_PATH],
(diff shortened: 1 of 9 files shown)


lib/ioki/apis/passenger_api.rb
```diff
@@ -113,6 +113,20 @@ class PassengerApi
         base_path:   [API_BASE_PATH],
         model_class: Ioki::Model::Passenger::Station
       ),
+      Endpoints::Create.new(
+        :logpay_customer,
+        base_path:            [API_BASE_PATH],
+        path:                 'logpay/customer',
```
rmehner, line 119: Same path thing as in the previous PR :)

lib/ioki/apis/passenger_api.rb
```diff
@@ -113,6 +113,20 @@ class PassengerApi
...
+        model_class:          Ioki::Model::Passenger::LogpayUrl,
+        outgoing_model_class: Ioki::Model::Passenger::LogpayCustomer
+      ),
+      Endpoints::Create.new(
+        :logpay_payment_method,
+        base_path:            [API_BASE_PATH],
+        path:                 'logpay/payment_method',
```
rmehner, line 126: See above :)

rmehner (approved):

lib/ioki/apis/passenger_api.rb
phylor: Changed!

lib/ioki/apis/passenger_api.rb
phylor: Changed!

lib/ioki/apis/passenger_api.rb
phylor: That doesn't work out of the box with `Endpoints::Create`. I'll merge it as is and create another PR for the consistency.

PR #108, opened 2022-06-30
Operator API | Add station endpoints and model

tom-ioki:
This PR introduces station endpoints and the station model for the operator API.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -38,6 +38,17 @@ module Ioki
         :place,
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::Place
+      ),
+      Endpoints.crud_endpoints(
+        :station,
+        base_path:   [API_BASE_PATH, 'products', :id],
+        model_class: Ioki::Model::Operator::Station
+      ),
+      Endpoints.custom_endpoints(
+        'stations',
+        actions:     { 'fix' => :patch, 'unfix' => :patch },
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -38,6 +38,17 @@ class OperatorApi
...
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::Place
+      ),
+      Endpoints.crud_endpoints(
+        :station,
+        base_path:   [API_BASE_PATH, 'products', :id],
+        model_class: Ioki::Model::Operator::Station
+      ),
+      Endpoints.custom_endpoints(
+        'stations',
```
rmehner, line 48: Any reason for not using a symbol here? Looks odd to me, since we're using symbols everywhere else, but I might miss something :)

tom-ioki: Hm no, not really. I was looking at the `platform_api` endpoints and found this
```ruby
      Endpoints.custom_endpoints(
        'drivers',
        actions:     { 'lock' => :patch, 'unlock' => :patch, 'set_pin' => :patch },
        path:        [API_BASE_PATH, 'products', :id, 'drivers', :id],
        model_class: Ioki::Model::Platform::Driver
      ),
```
So I thought I do the same thing. But I have no hard opinion on this. A symbol should also be fine. The `Endpoint` class will cast it to string anyways with this method:
```ruby
def name
      @method_name || "#{resource}_#{action}"
    end
```

ferdicus (approved):

PR #109, opened 2022-06-30
Platform API | Add reassign task list endpoint

ozb-ioki:
Add a new custom endpoint to reassign a new vehicle to an existing task list.

### DIFF
--- a/lib/ioki/apis/platform_api.rb
+++ b/lib/ioki/apis/platform_api.rb
@@ -123,6 +123,12 @@ module Ioki
         except:      [:create, :update, :delete],
         model_class: Ioki::Model::Platform::TaskList
       ),
+      Endpoints.custom_endpoints(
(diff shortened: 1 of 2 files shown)


lib/ioki/apis/platform_api.rb
```diff
@@ -123,6 +123,12 @@ class PlatformApi
         except:      [:create, :update, :delete],
         model_class: Ioki::Model::Platform::TaskList
       ),
+      Endpoints.custom_endpoints(
+        'task_list',
```
tom-ioki, line 127: Just some nit-picking here. I think we use the plural form of the model for custom endpoints. But I have no idea if this is really a rule or common practice 🤷 

No hard opinion here :)

tom-ioki (approved):
Nice work :)

PR #110, opened 2022-07-13
Passenger API | Add endpoints for service credits and personal discounts

phylor:
Allows to manage service credits and personal discounts.


rmehner (approved):

rmehner:
Should we create a new feature release for this?

phylor:
> Should we create a new feature release for this?

Sure, we can release a minor one. But maybe after all the logpay stuff has been merged. I also have one upcoming addition to delete payment methods we could also include.

PR #112, opened 2022-07-13
Operator API | Add task_list endpoints and model

tom-ioki:
This PR introduces task_list endpoints and the task_list model for the operator API.

This also solves #113 

I had to change some parts of the `Ioki::Model::Base` class in order to support multiple class names. Please have a thorough look there.


lib/ioki/model/operator/task_list.rb
```diff
@@ -4,28 +4,91 @@ module Ioki
...
+                  on:         [:create, :read, :update],
+                  type:       :object,
+                  class_name: %w[Place Station]
+
+        attribute :end_place,
+                  on:         [:create, :read, :update],
+                  type:       :object,
+                  class_name: 'Place'
+
+        # attribute :matching_configuration,
```
tom-ioki, line 49: For the moment this is not needed that's why I commented this part out.

lib/ioki/model/base.rb
```diff
@@ -174,6 +177,17 @@ def constantize_in_module(class_name)
 
         self.class.module_parent.const_get class_name
       end
+
+      def class_name_from_value_type(class_name, value)
+        return class_name unless class_name.is_a?(Array)
+        return nil if value.nil?
+
+        class_name_from_value_type = value['type'].split('_').map(&:capitalize).join('')
```
ans-ioki, line 185: I like this extension.. but i keep thinking about whether we need more "generic" specs for classes using these features...

Like a spec that uses a dummy class inheriting from model base and then using an array in this case here.

tom-ioki: Yes! That sounds like a good idea!

I was wondering how to add proper tests for this extension because the existing spec is running with vcr recorded examples I think. 🤔 

But I like the dummy class and inheriting from the base class idea more tbh. I will give it a try!

tom-ioki: **UPDATE**
I added some specs for the TaskList Model using a dummy class inheriting from the Ioki::Model::Base class.

eikes (commented):
I love the addition of the class name inference from the value type. I am wondering about the tests though. Why is the test about a fake class and not the actual TaskList class?

lib/ioki/model/base.rb
```diff
@@ -174,6 +177,23 @@ def constantize_in_module(class_name)
 
         self.class.module_parent.const_get class_name
       end
+
+      def class_name_from_value_type(class_name, value)
+        return class_name unless class_name.is_a?(Array)
+        return nil if value.nil?
+
+        class_name_from_value_type = value_type(value).split('_').map(&:capitalize).join('')
```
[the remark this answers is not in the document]
tom-ioki: Nice!

spec/helper/dummy_class_helpers.rb
```diff
@@ -0,0 +1,16 @@
+module Dummys
+end
+
+module DummyClassHelpers
+  def dummy_class(name, module_name, superclass = Object, &block)
```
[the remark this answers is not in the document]
tom-ioki: Deleted it 👍

spec/ioki/model/operator/task_list_spec.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+RSpec.describe Ioki::Model::Operator::TaskList do
+  let(:task_list) do
+    task_list_model.new(
```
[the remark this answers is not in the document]
tom-ioki: Changed it! 👍

eikes (approved):
Looks great!

PR #115, opened 2022-07-20
Passenger API | Available notification settings

phylor:
Adds endpoint to fetch available notification settings.

### DIFF
--- a/lib/ioki/apis/passenger_api.rb
+++ b/lib/ioki/apis/passenger_api.rb
@@ -98,7 +98,7 @@ module Ioki
       ),
       Endpoints::ShowSingular.new(
         :available_notification_settings,
-        path:        ['passenger', 'notification_settings/available'],
+        path:        %w[passenger notification_settings available],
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/passenger_api.rb
```diff
@@ -96,6 +96,12 @@ class PassengerApi
         base_path:   [API_BASE_PATH],
         model_class: Ioki::Model::Passenger::NotificationSettings
       ),
+      Endpoints::ShowSingular.new(
+        :available_notification_settings,
+        path:        ['passenger', 'notification_settings/available'],
```
rmehner, line 101: Very nitpicky, but shouldn't it be like this? (judging from other definitions)

```suggestion
        path:        ['passenger', 'notification_settings', 'available'],
```

rmehner (approved):

lib/ioki/apis/passenger_api.rb
phylor: Totally!

PR #118, opened 2022-08-16
Inherit rubocop config from Triebwerk

flp-ioki:
This is an effort to use the same coding style in all of our projects.

- See https://github.com/dbdrive/triebwerk/issues/10226 - if you can.


rmehner:
This is neat, I'll steal this and use it for Schalter as well, once this is done. Thanks @flp-ioki!

flp-ioki:
> This is neat, I'll steal this and use it for Schalter as well, once this is done. Thanks @flp-ioki!

Please do so! Then I don't have to do it. :smile: Also, you know much better which of the settings you already adjusted specifically for Schalter.

Thank you!

rmehner (approved):

Issue #97, opened 2022-05-20
Endpoints module should not be top level


eikes:
The `Endpoints` module should also be nested within the `Ioki` module

PR #120, opened 2022-08-26
Moves the 'Endpoints' module into the 'Ioki' module

eikes:
The Endpoints module shouldn't be a top level module. I assume it was be accident. This resolves the issue.


eikes:
See issue number #97

rmehner (approved):

rmehner:
Seems like an accident. Thanks for fixing!

PR #121, opened 2022-08-30
Operator API | Add fuel_type_options to Model::Operator::Vehicle

tom-ioki:
This adds the fuel_type_options attribute to the vehicle model of the operator api.

### DIFF
--- a/lib/ioki/model/operator/vehicle.rb
+++ b/lib/ioki/model/operator/vehicle.rb
@@ -44,6 +44,10 @@ module Ioki
                   omit_if_nil_on: [:create, :update],
                   type:           :string
 
+        attribute :fuel_type_options,
+                  on:   :read,
+                  type: :array
+
         attribute :last_known_position,
                   on:         :read,
                   type:       :object,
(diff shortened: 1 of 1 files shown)


lib/ioki/model/operator/vehicle.rb
```diff
@@ -44,6 +44,10 @@ class Vehicle < Base
                   omit_if_nil_on: [:create, :update],
                   type:           :string
 
+        attribute :fuel_type_options,
+                  on:   :read,
+                  type: :array
```
rmehner, line 49: Does this have a better type that we can describe with `classname` or is this just an arbitrary array of strings?

tom-ioki: The reason for adding this is to make sure that all fuel type options are included in a dropdown selection for the vehicle creation.

rmehner: Thanks for clarifying!

ans-ioki (approved):

PR #140, opened 2022-10-26
Ignore unrecognized rubocop cops

flp-ioki:
When Triebwerk updates its rubocop versions, it sometimes adds cops to the config files that older versions don't know about. [Rubocop recently added](https://github.com/rubocop/rubocop/pull/10630) the `--ignore-unrecognized-cops` flag for exactly this usecase.

I'm sorry for not figuring this out earlier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
           bundler-cache: true
       - name: Run rubocop
         run: |
-          bundle exec rubocop
+          bundle exec rubocop --ignore-unrecognized-cops
 
   backend-services:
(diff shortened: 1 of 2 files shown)


tom-ioki (approved):

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           bundler-cache: true
       - name: Run rubocop
         run: |
-          bundle exec rubocop
+          bundle exec rubocop --ignore-unrecognized-cops
```
rmehner, line 18: We don't need to add the CLI option here, as the config should be picked up automatically.

PR #141, opened 2022-10-26
Enable dependabot for this repo

rmehner:
This should allow us to notice when rubocop and friends update and keep
up with the upstream versions used at Triebwerk and Co.


tom-ioki (approved):
👍

PR #145, opened 2022-10-26
Operator Api | Add vehicle position endpoint

tom-ioki:
This PR adds the `create` endpoint for vehicle positions for the operator api.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -160,7 +160,7 @@ module Ioki
         except:      [:index, :show]
       ),
       Endpoints.crud_endpoints(
-        :vehicle_position,
+        :position,
         base_path:   [API_BASE_PATH, 'products', :id, 'vehicles', :id],
         model_class: Ioki::Model::Operator::VehiclePosition,
         except:      [:show, :index, :update, :delete]
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -158,6 +158,12 @@ module Ioki
         },
         model_class: Ioki::Model::Operator::Rating,
         except:      [:index, :show]
+      ),
+      Endpoints.crud_endpoints(
+        :vehicle_position,
+        base_path:   [API_BASE_PATH, 'products', :id, 'vehicles', :id],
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -158,6 +158,12 @@ class OperatorApi
         },
         model_class: Ioki::Model::Operator::Rating,
         except:      [:index, :show]
+      ),
+      Endpoints.crud_endpoints(
+        :position,
+        base_path:   [API_BASE_PATH, 'products', :id, 'vehicles', :id],
+        model_class: Ioki::Model::Operator::VehiclePosition,
```
rmehner, line 165: According to the docs, this is the Vehicles endpoint and not the position endpoint. And it returns a vehicle, not a position. I'm either very confused, or this is the wrong path :)

tom-ioki: Ah! The triebwerk PR which will introduce the vehicle position endpoint is not yet merged. That's why it is missing in the docs. Sorry 🙏 

The path will something like that:
`/api/operator/products/:product_id/vehicles/:vehicle_id/positions`

rmehner: That I assumed, but shouldn't the base_path above then reflect this? Like this:

```suggestion
        base_path:   [API_BASE_PATH, 'products', :id, 'vehicles', :id, 'positions'],
        model_class: Ioki::Model::Operator::VehiclePosition,
```

tom-ioki: The `crud_endpoints` method takes care of that. The resource (first parameter) `positions` will be attached to the base_path for the `create` action. Also have a look at the spec, there we check for the correct path :)

```ruby
it 'calls request on the client with expected params' do
      expect(operator_client).to receive(:request) do |params|
        expect(params[:url].to_s).to eq('operator/products/0815/vehicles/4711/positions')
        expect(params[:method]).to eq(:post)
        [result_with_data, full_response]
      end

      expect(operator_client.create_position('0815', '4711', vehicle_position, options))
        .to be_a(Ioki::Model::Operator::VehiclePosition)
    end
```

rmehner (approved):

PR #148, opened 2022-11-02
Operator Api | Add MatchingConfiguration CRUD endpoints

tom-ioki:
This PR will add CRUD endpoints for the MatchingConfiguration resource. For now, these endpoints will stay operator-api-only endpoints!

### DIFF
--- a/lib/ioki/model/operator/matching_configuration.rb
+++ b/lib/ioki/model/operator/matching_configuration.rb
@@ -26,8 +26,8 @@ module Ioki
                   class_name: 'Area'
 
         attribute :area_geojson,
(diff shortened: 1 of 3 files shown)


lib/ioki/model/operator/matching_configuration.rb
```diff
@@ -25,13 +25,150 @@ class MatchingConfiguration < Base
                   type:       :object,
                   class_name: 'Area'
 
+        attribute :area_geojson,
```
tom-ioki, line 28: This should be of type: :string

ferdicus (approved):

PR #154, opened 2022-11-17
General | Enhance Error Handling

tom-ioki:
At the moment, ioki-ruby is not able to handle errors that have a non-json response body (e.g. some errors like `404` - `ActiveRecord::RecordNotFound` come with a 'text/html' response body). Reason being, the `JsonParse` middleware of Faraday tries to parse the response body to json no matter what. This however causes ioki-ruby to raise a `Faraday::ParsingError` when the response body is not a json.

This PR will add the `content_typ` setting to the JsonParse middleware of Faraday which ensures that only response bodies of type json will be parsed. Response bodies of any other type will not be parsed and thus will not raise the infamous `Faraday::ParsingError`.

### DIFF
--- a/lib/ioki/http_adapter.rb
+++ b/lib/ioki/http_adapter.rb
@@ -24,8 +24,10 @@ module Ioki
             logger.filter(/("token"=>")([^"]+)/, '\1[REMOVED]')
           end
         end
-        f.response :json # decode response bodies as JSON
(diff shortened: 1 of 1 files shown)


rmehner (approved):

rmehner:
Neat. We should put out a new release soon :)

PR #161, opened 2022-12-20
Platform API | Add endpoints for captchas

phylor:
This adds required endpoints to implement captchas on the platform API, namely:

* Post a captcha solution `/api/platform/captchas/{id}/solution`
* Regenerate the captchas task `/api/platform/captchas/{id}/regenerate`

It also adds the captcha model and adds it to the response of the endpoint "Create a new Phone Verification Request" `/api/platform/providers/{provider_id}/phone_verification_requests`

### DIFF
--- a/lib/ioki/model/platform/captcha.rb
+++ b/lib/ioki/model/platform/captcha.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Platform
+      class Captcha < Base
+        attribute :type, on: :read, type: :string
+        attribute :id, on: :read, type: :string
+        attribute :created_at, on: :read, type: :date_time
+        attribute :updated_at, on: :read, type: :date_time
+        attribute :captcha_type, on: :read, type: :string
+        attribute :question_prompt, on: :read, type: :string
+        attribute :image_url, on: :read, type: :string
+
+        attribute :solution, on: :create, type: :string
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


rmehner:
@phylor is this ready for review?

phylor:
> @phylor is this ready for review?

Yes!

lib/ioki/model/platform/captcha.rb
```diff
@@ -0,0 +1,19 @@
...
+    module Platform
+      class Captcha < Base
+        attribute :type, on: :read, type: :string
+        attribute :id, on: :read, type: :string
+        attribute :created_at, on: :read, type: :date_time
+        attribute :updated_at, on: :read, type: :date_time
+        attribute :captcha_type, on: :read, type: :string
+        attribute :question_prompt, on: :read, type: :string
+        attribute :image_url, on: :read, type: :string
+
```
rmehner, line 14: 🤓 Nitpick, but what's this blank line doing? :D

rmehner (approved):

lib/ioki/model/platform/captcha.rb
phylor: It is feng shui of coding. It let's the energy flow freely through the code :cowboy_hat_face: 

In seriousness, all the top attributes are `on: :read` and the last one is `on: :create`, so this visually groups them. I usually like to have some grouping in such lists for better understanding, but not sure if this is the right one in this case.

rmehner: Neat, didn't see that. Makes sense.

PR #171, opened 2023-01-05
Platform Api | Refine UserEmail

tom-ioki:
This PR adds two new attributes (`confirmed_email_address` and `unconfirmed_email_address`) to the UserEmail model for the platform api.

This means, when creating a new user a confirmed or an unconfirmed email address can be send to triebwerk. The attribute `email_address` can still be used but is deprecated now.

### DIFF
--- a/lib/ioki/apis/platform_api.rb
+++ b/lib/ioki/apis/platform_api.rb
@@ -44,6 +44,12 @@ module Ioki
         base_path:   [API_BASE_PATH, 'providers', :id],
         model_class: Ioki::Model::Platform::User
       ),
+      Endpoints.custom_endpoints(
+        'user',
+        actions:     { 'email' => :delete },
+        path:        [API_BASE_PATH, 'providers', :id, 'users', :id],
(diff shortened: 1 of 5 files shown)


rmehner:
@tom-ioki this is live on Triebwerk right? So after this change, we could go to Schalter and implement this to solve the deprecation?

rmehner (approved):

rmehner:
Likely needs an update to https://github.com/ioki-mobility/ioki-ruby/blob/a633733c04c1b2801473edb9d63fca459ceb4e1b/spec/ioki/model/platform/user_email_spec.rb

rmehner:
Is this the same for Passenger, or only Platform?

tom-ioki:
> @tom-ioki this is live on Triebwerk right? So after this change, we could go to Schalter and implement this to solve the deprecation?

Sorry for the late response!
No, this change is not live yet. But will be in the next days I guess 👍 

And yes, you can solve the deprecation then 👍

PR #181, opened 2023-01-27
Change notification settings to new API

phylor:
There has been some changes to the notification settings API. This PR accomodates for that.

The old API returned an object, such as:

```ruby
{
  data: {
    ride_notifications: ['sms']
  }
}
```

Now, the API returns an array of `NotificationSetting`:

```ruby
{
  data: [
    {
      id: 'ride_notifications',
      name: 'ride_notifications',
      channels: ['sms']
    }
  ]
}
```

### DIFF
--- a/lib/ioki/apis/endpoints/show.rb
+++ b/lib/ioki/apis/endpoints/show.rb
@@ -28,7 +28,12 @@ module Ioki
         # :show can optionally use the model for caching.
         model = options[:model] if options[:model].is_a?(model_class)
         attributes, etag = model_params(client, args, options, model)
-        model_class.new(attributes, etag)
+
+        if model_class == Array
+          attributes
+        else
+          model_class.new(attributes, etag)
+        end
       end
 
       private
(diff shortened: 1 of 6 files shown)


phylor:
@tom-ioki It would be great to get your opinion on this PR. I had to hack around a bit to allow an API response to be an `Array` instead of a `Hash`. It works, but it looks hacky and it doesn't support etags when an array is returned.

tom-ioki:
> @tom-ioki It would be great to get your opinion on this PR. I had to hack around a bit to allow an API response to be an `Array` instead of a `Hash`. It works, but it looks hacky and it doesn't support etags when an array is returned.

Hey 👋 
sorry for the late reply. It took me a while to understand the issue. Maybe we should have a short call?

tom-ioki (approved):
Looks good to me 👍 

As discussed, I will create a follow up issue for this. Maybe we can move this logic into the model world and use some kind of keyword which indicates that the response will be an array of hashes or something like that.

PR #184, opened 2023-01-31
Add support for bodies other than objects

### LINKED ISSUE #183: Handle array responses
## Problem
With the current changes for `NotificationSettings` the response from triebwerk changed from a hash

```ruby
{
  data: {
    ride_notifications: ['sms']
  }
}
``` 

to an array of hashes (NotificationSetting objects)

```ruby
{
  data: [
    {
      id: 'ride_notifications',
      name: 'ride_notifications',
      channels: ['sms']
    }
  ]
}
```

The current solution makes handling these responses possible but does not support etags.

## Solution
Let's find a better place for the array handling within the gem. Maybe we can implement the array handling in the `Ioki::Model::Base` class and add some kind of keyword that indicates that the incoming data will be an array of hashes.

---
Follow up issue for this PR https://github.com/ioki-mobility/ioki-ruby/pull/181.

phylor:
Closes #183.

## TLDR

This adds support for bodies we receive, which are not an object. For example, arrays or plain strings.

Let's say we get the response:

```ruby
{
  data: [
    { name: 'ride_notifications', channels: ['sms'], type: 'notification_setting' }
  ]
}
```

You could now parse this with the following model:

```ruby
class Ioki::Model::Passenger::NotificationSettings
  base 'Array', item_class_name: 'NotificationSetting'
end
```

## Changes

* Up until now all response bodies we parse in ioki-ruby need to be objects (in practice a `Hash`). We now support arrays and basic types (`String`, ..) as well.

  For arrays we need to define an `item_class_name` which specifies the class of the items in the array.
  
  ```ruby
  class Ioki::Model::Passenger::NotificationSettings
    base 'Array', item_class_name: 'NotificationSetting'
  end
  ```
  
  For basic types, we only define the type:
  
  ```ruby
  class Ioki::Model::Passenger::ShuttleColor
    base 'String'
  end
  ```

* Note that even if the response is a basic type or an array, ioki-ruby will always return a model inheriting `Ioki::Model::Base`. This allows for helper methods such as `serialize` and also allows access to the etag using `model._etag`.

  ```ruby
  model = Ioki::Model::Passenger::NotificationSettings.new([{name: 'ride', channels: ['sms']}], 'etag-123')
  
  model._etag
  # => 'etag-123'
  
  model._raw_attributes
  # => [{name: 'ride', channels: ['sms']}]
  # `#_raw_attributes` returns the input value
  
  model.data
  # => [Ioki::Model::Passenger::NotificationSetting<name='ride' channels=['sms']>]
  # `#data` returns the parsed objects
  ```

* New method `#data` which returns the parsed data. For object responses, it returns `#attributes`. The `#data` method should be used in client applications to get the parsed array response.

  ```ruby
  model = Ioki::Model::Passenger::Product.new(name: 'MyProduct')

  model.attributes
  # => {name: 'MyProduct'}

  model.data
  # => {name: 'MyProduct'}
  ```

### DIFF
--- a/lib/ioki/apis/endpoints/show.rb
+++ b/lib/ioki/apis/endpoints/show.rb
@@ -29,11 +29,7 @@ module Ioki
         model = options[:model] if options[:model].is_a?(model_class)
         attributes, etag = model_params(client, args, options, model)
 
-        if model_class == Array
-          attributes
-        else
-          model_class.new(attributes, etag)
-        end
+        model_class.new(attributes, etag)
       end
 
       private
--- a/lib/ioki/apis/passenger_api.rb
+++ b/lib/ioki/apis/passenger_api.rb
@@ -93,19 +93,19 @@ module Ioki
         :notification_settings,
         path:        'notification_settings',
         base_path:   [API_BASE_PATH],
-        model_class: Array
+        model_class: Ioki::Model::Passenger::NotificationSettings
       ),
       Endpoints::ShowSingular.new(
         :default_notification_settings,
         path:        %w[passenger notification_settings defaults],
         base_path:   [API_BASE_PATH],
-        model_class: Array
+        model_class: Ioki::Model::Passenger::NotificationSettings
       ),
       Endpoints::ShowSingular.new(
         :available_notification_settings,
         path:        %w[passenger notification_settings available],
         base_path:   [API_BASE_PATH],
-        model_class: Array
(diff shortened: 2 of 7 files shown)


phylor:
@tom-ioki If you have time, this would be a PR about array bodies. Now supporting etags. Let me know if there are things you imagined differently!

tom-ioki (approved):
Great work! Love it and thank you 😊

PR #186, opened 2023-02-03
Operator Api | Add external_id to Driver model

tom-ioki:
This PR adds some missing attributes to the Driver model.


ferdicus (approved):
👍🏿

PR #187, opened 2023-02-06
PassengerAPI | Fix updating language

phylor:
Currently, the `Passenger#update_language` endpoint doesn't work as expected.

What should happen:

1. Call `my_passenger_client.update_language('fr')`
2. ioki-ruby calls `PATCH /api/passenger/user/language` and sets the `Accept-Language` header to `fr`. That header is used to set the language. There is no body or path parameters.

What actually happens:

1. ioki-ruby is initialized with a `Ioki::Configuration`. At that point the language is taken from the configuration and an HTTP adapter is initialized as `config.http_adapter` with that language.
2. Call `my_passenger_client.update_language('fr')`
3. The method sets `config.language` to the new language and then makes the HTTP request
4. However, the new config has no effect, because the HTTP adapter has been initialised with the initial configuration.

So we have to re-initialize the HTTP adapter, so that the new language takes effect and is sent as the HTTP header.

### DIFF
--- a/lib/ioki/apis/endpoints/passenger/update_language.rb
+++ b/lib/ioki/apis/endpoints/passenger/update_language.rb
@@ -10,6 +10,7 @@ module Ioki
 
         def call(client, args = [], options = {})
           client.config.language = args.last
+          client.config.http_adapter = Ioki::HttpAdapter.get(client.config) unless client.config.custom_http_adapter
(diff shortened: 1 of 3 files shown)


tom-ioki (approved):
That sounds annoying to debug. Nice fix 👍 Thanks

PR #209, opened 2023-04-11
Let user provide the signature_key to Ioki::Webhooks::SignatureValidator

eikes:
There is no need to define how the signature key is set. In our case the ENV is a bad choice. This PR fixes that...

### DIFF
--- a/lib/ioki/webhooks/signature_validator.rb
+++ b/lib/ioki/webhooks/signature_validator.rb
@@ -18,25 +18,28 @@ require 'openssl'
 module Ioki
   module Webhooks
     class SignatureValidator
-      def initialize(body:, signature:)
+      def initialize(body:, signature:, signature_key:)
         @body = body
         @signature = signature
+        @signature_key = signature_key
       end
 
       def call
         raise Ioki::Error::WebhookSignatureMissing if Ioki::Support.blank? @signature
         raise Ioki::Error::WebhookBodyMissing if Ioki::Support.blank? @body
-        raise Ioki::Error::WebhookSignatureKeyMissing if Ioki::Support.blank? ENV.fetch('WEBHOOK_SIGNATURE_KEY', nil)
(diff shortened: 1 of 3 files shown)


lib/ioki/webhooks/signature_validator.rb
```diff
@@ -18,25 +18,28 @@
 module Ioki
   module Webhooks
     class SignatureValidator
-      def initialize(body:, signature:)
+      def initialize(body:, signature:, signature_key:)
```
rmehner, line 21: This is a breaking change. We don't have a proper release process yet to highlight this (although we do have a version file, so we just at least bump that).

However, I wonder if we should go for this for now:

```suggestion
      def initialize(body:, signature:, signature_key: ENV.fetch('WEBHOOK_SIGNATURE_KEY', nil))
```

This would allow dependent projects to do the upgrade and once we have a release process in place, we could easily increase the major version with proper release notes.

Or do you reckon it's fine as is, as this would throw an ArgumentError if you just update the dependency without doing the necessary call update and rely on the respective test suites to catch that?

rmehner (approved):
Change in general is good and explicit > implicit. However, I'd like to discuss a sensible way to release this (see comment)

lib/ioki/webhooks/signature_validator.rb
eikes: Fair point! Indeed it would throw an error and we should mention it in the release notes. But I like your idea of providing  the ENV variable by default, it is the most seamless transition.

eikes:
Should we create a new release?

rmehner:
> Should we create a new release?

We don't have the concept of a release, but I'd at least bump the version in https://github.com/ioki-mobility/ioki-ruby/blob/main/lib/ioki/version.rb to kind of show that there's an important change.

I'd love to have a proper release process & changelog for this.

PR #213, opened 2023-04-17
Add support for bleeding edge API

phylor:
This adds support for the "bleeding edge" API version.

It defaults to `false` by default. You can either configure it using `Ioki.config.api_bleeding_edge` or using the environment variable `IOKI_API_BLEEDING_EDGE`.

### DIFF
--- a/spec/ioki/http_adapter_spec.rb
+++ b/spec/ioki/http_adapter_spec.rb
@@ -15,6 +15,7 @@ RSpec.describe Ioki::HttpAdapter do
       api_client_secret:     SecureRandom.alphanumeric,
       api_client_version:    'api_client_version',
       api_token:             SecureRandom.alphanumeric,
+      api_bleeding_edge:     false,
       language:              :'en-BZ',
       logger:                logger,
       logger_options:        { headers: true, bodies: true }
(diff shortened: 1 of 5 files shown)


spec/ioki/http_adapter_spec.rb
```diff
@@ -68,6 +70,14 @@
...
       it { is_expected.not_to include 'X-Client-Identifier' }
     end
+
+    context 'when bleeding edge is not set' do
+      before do
+        config.api_bleeding_edge = false
+      end
+
+      it { is_expected.not_to include 'X-Bleeding-Edge' => config.api_bleeding_edge }
+    end
```
rmehner, line 80: What's the reasoning for testing the default with a special test, while testing the special case as default? I'd expect it to be the other way around (granted, this is a bit of a nitpick)

rmehner (approved):

spec/ioki/http_adapter_spec.rb
phylor: I was thinking too much about my own use case :grinning: I agree, will switch it!

spec/ioki/http_adapter_spec.rb
```diff
@@ -15,6 +15,7 @@
       api_client_secret:     SecureRandom.alphanumeric,
       api_client_version:    'api_client_version',
       api_token:             SecureRandom.alphanumeric,
+      api_bleeding_edge:     false,
```
rmehner, line 18: I'd actually not have this line here, since do want to test the default, right?

phylor: Yes! See #215.

PR #217, opened 2023-04-25
Operator Api | Add more attributes to Place model

tom-ioki:
This will add `depot` and `waiting_position` to the place model for the operator api.

### DIFF
--- a/lib/ioki/model/operator/place.rb
+++ b/lib/ioki/model/operator/place.rb
@@ -73,9 +73,9 @@ module Ioki
                   on:   [:create, :read, :update],
                   type: :string
 
-        atribute :waiting_position,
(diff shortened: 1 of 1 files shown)


lib/ioki/model/operator/place.rb
```diff
@@ -32,6 +32,10 @@ class Place < Base
                   on:   [:read, :create, :update],
                   type: :string
 
+        attribute :depot,
+                  on:   [:read, :create, :update],
```
ferdicus, line 36: sometimes it's `[:read, :create, :update]` and other times `[:create, :read, :update]` - why you not align them alphabetically? 🤷🏿 

sorry, i wanted to at least once comment on a PR 😄

ferdicus (approved):

PR #220, opened 2023-05-09
fix: place model attributes

tom-ioki:
The `depot` and `waiting_position` attribute should be omitted if their value is nil in incoming requests.


ferdicus (approved):
lgtm

PR #226, opened 2023-05-23
Setup omit_if_nil_on to version

tom-ioki:
Apparently, updating some resources result in 422 at the moment. Reason being that when no value for version is set `nil` will be send in to triebwerk which produces problems.


ferdicus (approved):
LGTM 👍🏿

PR #227, opened 2023-05-25
Operator Api | Add ride inquiry endpoint

tom-ioki:
This PR will add the ride inquiry endpoint for the operator api.

### DIFF
--- a/lib/ioki/model/operator/constraints.rb
+++ b/lib/ioki/model/operator/constraints.rb
@@ -6,22 +6,6 @@ module Ioki
       class Constraints < Base
         unvalidated true
 
-        # attribute :max_passengers,
-        #           on:   :read,
-        #           type: :integer
-
-        # attribute :max_wheelchairs,
-        #           on:   :read,
-        #           type: :integer
-
-        # attribute :max_walkers,
-        #           on:   :read,
-        #           type: :integer
-
-        # attribute :max_storage_spaces,
-        #           on:   :read,
-        #           type: :integer
-
         attribute :area,
                   on:         :read,
                   type:       :object,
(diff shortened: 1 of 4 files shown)


lib/ioki/model/operator/constraints.rb
```diff
@@ -6,11 +6,31 @@ module Operator
...
+        #           on:   :read,
+        #           type: :integer
+
+        # attribute :max_walkers,
+        #           on:   :read,
+        #           type: :integer
+
+        # attribute :max_storage_spaces,
+        #           on:   :read,
+        #           type: :integer
```
rmehner, line 23: Why are these attributes commented out? If we don't need them, we can remove them and git can bring them back if we need to :)

lib/ioki/model/operator/ride_inquiry.rb
```diff
@@ -4,19 +4,61 @@ module Ioki
...
+
+        # attribute :assistances,
+        #           on:         :read,
+        #           type:       :array,
+        #           class_name: 'Assistance'
+
+        # attribute :availability,
+        #           on:         :read,
+        #           type:       :object,
+        #           class_name: 'Availability'
```
rmehner, line 47: Same here, why are those commented out?

rmehner (approved):

ans-ioki:
Same here - i'm with robin, no commented out code.. otherwise fine!

lib/ioki/model/operator/constraints.rb
tom-ioki: Yes, you're right. I kept them in case something changes. But after some testing I can confirm that these attributes are not needed at the moment.

lib/ioki/model/operator/ride_inquiry.rb
tom-ioki: I will remove them. Thanks for the review 👍

lib/ioki/model/operator/constraints.rb
tom-ioki: done!

PR #243, opened 2023-07-11
Operator Api | Add product default attributes to Product model

tom-ioki:
This PR will add product default attributes to the `Product` model for the operator api.


ferdicus (approved):
thanks

PR #246, opened 2023-07-18
Operator Api | Add `matching_rank` to Operator, Vehicle and TaskList model

tom-ioki:
This PR will add the attribute `matching_rank` to the Operator, Vehicle and TaskList model for the operator api.


ferdicus (approved):
thx

PR #247, opened 2023-07-19
Operator/Platform API | Add update of line stops

phylor:
Adds updating of line stops to the operator and platform API.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -194,7 +194,7 @@ module Ioki
       Endpoints.crud_endpoints(
         :line_stop,
         base_path:   [API_BASE_PATH, 'products', :id, 'lines', :id],
-        except:      [:create, :delete],
+        except:      [:index, :show],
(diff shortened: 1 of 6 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -194,7 +194,7 @@ class OperatorApi
       Endpoints.crud_endpoints(
         :line_stop,
         base_path:   [API_BASE_PATH, 'products', :id, 'lines', :id],
-        except:      [:create, :delete],
+        except:      [:index, :show],
```
rmehner, line 197: This change is weird, why do we need it to support update?

phylor: Yes, I believe we don't. But wasn't it completely wrong before? It's called `except`, but we included those actions we actually wanted.

rmehner: Makes sense, but it wasn't a separate commit, so I was wondering what the reason for this would be. Thanks for clarifying :)

rmehner (approved):

PR #251, opened 2023-08-02
Operator API | Add creation of repositioning tasks

phylor:
Adds the creation of repositioning tasks to the operator API.


tom-ioki (approved):
LGTM

PR #253, opened 2023-08-07
Platform API | Change to remove newsletter and receipt from user email if not provided

phylor:
If newsletter and receipt are set as `nil` the API returns a validation error that they are not booleans. We therefore omit them if they are not provided.


tom-ioki (approved):
Thanks for fixing this 👍

PR #257, opened 2023-08-11
Operator Api: add admin_sign_out endpoint for the operator api

tom-ioki:
This adds an endpoint for signing out the admin via the operator api.


tom-ioki:
@phylor is this PR useful for the Global Logout effort or should I close it?

PR #258, opened 2023-08-17
Refresh OAuth token and retry requests

phylor:
## Changes

* Adds a transparent OAuth token refresh. Only active when the client is initialized with OAuth settings
* Adds an automatic retry mechanism if a request fails (if we have an HTTP response, only for `5xx` errors)
* Fix warning when running rspec:

  ```
  WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives,
  since literally any other error would cause the expectation to pass, including those raised 
  by Ruby (e.g. `NoMethodError`, `NameError` and `ArgumentError`), meaning the code you
  are intending to test may not even get reached. Instead consider using `expect { }.not_to       
  raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`. This message can 
  be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`.
  ```

## Todos

- [x] Only retry on `5xx` errors


tom-ioki:
Hey👋 This looks promising.
I was testing this branch for Stellwerk. Unfortunately, I was getting a `NameError: uninitialized constant Ioki::Retry` error. Do I miss something in the config?

phylor:
@tom-ioki `require` fixed and now HTTP requests are only retried if they are a #500. Please have another look/try!

tom-ioki (approved):
Sorry for the late review!

Looks good to me. Also, I did some testing for Stellwerk and as far as I can tell everything is working. Finally, I can get rid of the `retry` class :D

PR #264, opened 2023-08-30
Passenger/Platform API | Dynamic passenger/ride options

phylor:
This adds the options for the dynamic passenger/ride options to the passenger and platform API.

### DIFF
--- a/lib/ioki/model/platform/ride_option.rb
+++ b/lib/ioki/model/platform/ride_option.rb
@@ -3,8 +3,9 @@
 module Ioki
   module Model
     module Platform
-      class PassengerType < Base
-        unvalidated true # Specification not available.
+      class RideOption < Base
+        attribute :name, on: [:read, :create], type: :string
+        attribute :value, on: :create, type: :string # could also be a boolean or an integer
 
         attribute :type, on: :read, type: :string
         attribute :id, on: :read, type: :string
@@ -12,8 +13,7 @@ module Ioki
         attribute :updated_at, on: :read, type: :date_time
 
         attribute :bookable, on: :read, type: :boolean
-        attribute :name, on: :read, type: :string
-        attribute :slug, on: :read, type: :string
+        attribute :data_type, on: :read, type: :string
         attribute :option_type, on: :read, type: :string
         attribute :name_translations, on: :read, type: :object, class_name: 'MultilanguageString'
         attribute :description_translations, on: :read, type: :object, class_name: 'MultilanguageString'
--- a/lib/ioki/model/platform/ride_option.rb
+++ b/lib/ioki/model/platform/ride_option.rb
@@ -5,7 +5,7 @@ module Ioki
     module Platform
       class RideOption < Base
         attribute :name, on: [:read, :create], type: :string
-        attribute :value, on: :create, type: :string # could also be a boolean or an integer
+        attribute :value, on: :create, type: [:string, :boolean, :integer]
 
         attribute :type, on: :read, type: :string
         attribute :id, on: :read, type: :string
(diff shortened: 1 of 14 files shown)


lib/ioki/model/platform/ride_option.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Platform
+      class RideOption < Base
+        attribute :name, on: [:read, :create], type: :string
+        attribute :value, on: :create, type: :string # could also be a boolean or an integer
```
rmehner, line 8: I briefly looked into supporting more that one type and it's a little bit more involved. Just feels weird that we might pass a boolean here and have it sent as string. Can the API handle that?

We'd need to adapt [type_cast_attribute_value](https://github.com/ioki-mobility/ioki-ruby/blob/dffab1bdfe37e136ce9728dfd8075d6ab8ed69bb/lib/ioki/model/base.rb#L124-L167) to work with an array of possible types and then check the incoming value against those types. 

Is that worth it?

rmehner (commented):
Generally looks good, just missing some specs :)

lib/ioki/model/platform/ride_option.rb
phylor: > I briefly looked into supporting more that one type and it's a little bit more involved. Just feels weird that we might pass a boolean here and have it sent as string. Can the API handle that?

No, I don't think so. The API is quite picky about the right data types 😀 I only tested with strings so far.

I added support for it in b9f9212e75a4051dbd5cd8347f0ac687f31828dd.

phylor:
> Generally looks good, just missing some specs :)

I added specs for the multi-type attributes. Would you expect other specs?

tom-ioki (approved):
Looks good to me and sounds like a reasonable idea 👍

tom-ioki:
Thanks for the detailed information 🙏

rmehner:
> If there is no matched type, we return `nil`. We could also return the parsed value (although the type doesn't match) or raise an exception.

I'm wondering what you'd expect here as a user of the client.

1. Exception might be way too harsh, because if Triebwerk ever changes something, we don't want clients to fail hard. Although it would help getting things fixed quickly 😄 
2. Returning `nil` would mean that the value is lost and it might lead to a subtle bug if someone expects something else. It would allow for proper nil checking on consumer side though, which might be a good pattern to have in consuming apps.
3. Returning the value parsed as is seems to be the best variant to me, as this would allow consumers to work with those values, even if ioki-ruby wasn't updated yet. Downside is that this doesn't put pressure on ioki-ruby to be up to date and we devs love weird workarounds :0

So I'm slightly in favor of 3, but 2 would also be okay with me.

phylor:
> 2. Returning `nil` would mean that the value is lost and it might lead to a subtle bug if someone expects something else. It would allow for proper nil checking on consumer side though, which might be a good pattern to have in consuming apps.
> 3. Returning the value parsed as is seems to be the best variant to me, as this would allow consumers to work with those values, even if ioki-ruby wasn't updated yet. Downside is that this doesn't put pressure on ioki-ruby to be up to date and we devs love weird workarounds :0
> 
> So I'm slightly in favor of 3, but 2 would also be okay with me.

Hmm, I'm a bit indifferent here :slightly_smiling_face: I generally prefer "fail early", so I'd prefer to return `nil` instead of a wrong type.

I'll go with the path of least resistance and merge as is (with option 2).

@rmehner What do you think: we change it to option 3 (returning whatever value we got) and then we additionally do

```ruby
STDERR.puts "We expected `Ioki::Model::Passenger::LineStop::supports_pickup` to be a `boolean`, but got 'i_am_a_string'"
```

PR #274, opened 2023-09-19
Operator Api | User autocomplete

tom-ioki:
This PR will add the endpoint `autocomplete` for the user resource.


pak-ioki (commented):
Why don't you just use the user model? I don't see any nested attributes there.

tom-ioki:
> Why don't you just use the user model? I don't see any nested attributes there.

That's a great point! The `UserAutocomplete` model looked a bit different in the beginning but with changing the attributes it's just a subset of the `User` model (except for the `full_name` but I could add this to the user serializer in Triebwerk). I will change this. Thank you for pointing this out 👍

tom-ioki:
@pak-ioki I've removed the `UserAutocomplete` model and changed the resulting class to the `User` model for the autocomplete endpoint. Also I've added some specs for the User model 👍

pak-ioki (approved):
LGTM

PR #277, opened 2023-09-25
Add matching modes to line stops

phylor:
This adds the matching modes `dropoff_mode` and `pickup_mode` to the line stops.


tom-ioki (approved):
👍

PR #278, opened 2023-09-27
Add `omit_if_nil_on` to matching mode attributes

phylor:
Because `dropoff_mode` and `pickup_mode` are optional attributes, we should not send them if they are `nil`.


tom-ioki (approved):
Alright 👍

PR #280, opened 2023-09-28
Operator Api | Add `User` to Ride model

tom-ioki:
This will add the user to the ride model.


pak-ioki (changes requested):
I miss the spec

tom-ioki:
> I miss the spec

Usually we don't need specs for the models, but I added some 👍

pak-ioki (approved):
Now it looks good to me

PR #284, opened 2023-09-29
Removes duplicate attribute definitions

eikes:


lib/ioki/model/passenger/user.rb
```diff
@@ -27,7 +27,6 @@ class User < Base
         attribute :registered, on: :read, type: :boolean
         attribute :remaining_referrals, on: :read, type: :integer
         attribute :unique_customer_id, on: :read, type: :string
-        attribute :logpay_customer_set, on: :read, type: :boolean
```
phylor, line 30: @eikes We still need this attribute in an app. It is also in the API documentation.

Did you remove it because it's not in the OpenAPI spec? If so, we probably should add it there.

eis-ioki: I removed it here because it is defined twice. All the removals in this PR are because the same attribute was defined twice in the model. It's still there. See for yourself: https://github.com/ioki-mobility/ioki-ruby/blob/e2fe57720ee07bbc32ade1174b6ea99696ac6ad6/lib/ioki/model/passenger/user.rb

lib/ioki/model/operator/payment_method.rb
```diff
@@ -5,10 +5,9 @@ module Model
     module Operator
       class PaymentMethod < Base
         attribute :type, on: :read, type: :string
-        attribute :id, on: :read, type: :string
+        attribute :id, on: [:read, :create], type: :string, omit_if_blank_on: :create
```
tom-ioki, line 8: That's kinda odd. The `id` should not be accessible via `create` because it is set by triebwerk. Why was it configured like that in the first place 🤔

lib/ioki/model/passenger/user.rb
phylor: Oh, so I added it despite it being there already :thinking: Sorry for the noise!

lib/ioki/model/operator/payment_method.rb
eikes: That makes sense. I'll remove that `:create` entirely.

tom-ioki: thanks 🙏

eikes: So I think I know what's going on here: When the `PaymentMethod` is nested within another object, it already has an `:id`, so when that other object is created, the `:id` should be there. When a `PaymentMethod` is created though, it doesn't have an `:id` so we should omit it. Therefor, all these `attribute :id, on: [:read, :create], type: :string, omit_if_blank_on: :create` do make sense...

phylor: Right, for example in `/api/passenger/rides/{id}/booking`. Thanks for catching it! We would probably have complained as soon as our tests began to fail :slightly_smiling_face:

eis-ioki: @phylor I added it back in and fixed a failing test. Could you please take a look now?

tom-ioki: Although, I guess it doesn't really make sense for the operator api 🤷

eikes: So it's a copy and paste result? Should we remove it then?

tom-ioki: > So it's a copy and paste result?
Most probably. 
> Should we remove it then?
Not sure 😄

tom-ioki: I guess it's fine to leave it for now 👍 Thanks @eikes

phylor: @eikes Our tests look fine! :+1:

tom-ioki (approved):
LGTM and thanks. This helps keeping the models a bit cleaner.

PR #285, opened 2023-09-29
Update ruby to last maintained ruby version

eikes:
Ruby 2.7.4 is not maintained anymore. We should raise the ruby version to 3.

https://www.ruby-lang.org/en/downloads/branches/


tom-ioki (approved):
Looks good from the operator tool perspective. @phylor any objections from your side?

phylor:
Looks good for our apps. Thank you!

PR #288, opened 2023-10-03
Passenger API | Payment details

phylor:
This changes:

- Add `stripe_payment_method_id` to be able to create stripe payment methods
- Change other payment method details not to be sent when `nil`. Otherwise the server side validation fails.


tom-ioki (approved):
LGTM

PR #296, opened 2023-10-26
Mailer

phylor:


README.md
```diff
@@ -241,3 +241,34 @@ If you would like to make those request with fast predictable results in you tes
...
+```ruby
+class MyMailer < ApplicationMailer
+  def welcome_message
+    mail(
+      subject: 'Welcome to our app!'
+    ).tap do |message|
+      message.ioki_options = {
+        provider_id:      my_provider_api_id,       # required
+        platform_client:  my_ioki_platform_client,  # required
+        delivery_context: 'standard',               # optional
```
ans-ioki, line 268: Lets make this required - depending on how you evolve the Triebwerk PR

lib/ioki/mailing/mailer.rb
```diff
@@ -0,0 +1,48 @@
...
+            mail(subject: 'Test E-Mail').tap do |message|
+              message.ioki_options = {
+                provider: my_provider,
+                platform_client: Ioki::Client.new(Ioki::Configuration.new(my_ioki_config), Ioki::PlatformApi)
+              }
+            end
+          ERROR
+        end
+
+        delivery_context = mail.ioki_options[:delivery_context] || 'standard'
```
ans-ioki, line 26: Same here, let's make this a required option

ans-ioki (approved):
Awesome!

README.md
phylor: `delivery_context` is required by the API, but I thought it would make sense to fallback to the standard option when none is passed. So the API always receives a value, but the Ruby caller doesn't need to provide one. It's one less option when sending an email. Would you still prefer that the caller always needs to pass the option to make sure it's the correct value?

PR #300, opened 2023-10-30
Remove the `operator_id` from update

phylor:
The operator can only be set when creating a vehicle, but not when updating. This is valid for the operator and the platform API.


tom-ioki (approved):
Didn't know about this change 😱 Interesting! And thanks for taking care of this.

PR #302, opened 2023-11-09
CRUD endpoints | Fix `except` option and add `only` option

phylor:
This fixes the `except` option when passed to `Endpoints.crud_endpoints`. The problem was that we didn't correctly deduce the action name (e.g. `:index`) from the action classes (e.g. `Ioki::Endpoints::Index`).

Mainly because `type` was e.g. `Ioki::Endpoints::Index`, but then calling `type.class` results in `Class` (because `type` is already a constant). So we checked whether the passed in *symbols* are equal to the string `"class"`, which never was true, hence the `except` option didn't have any effect.

Additionally, this PR adds an `only` option. If you only need 2 actions, it's easier to specify it as `only` instead of `except`.

### DIFF
--- a/lib/ioki/apis/endpoints/endpoints.rb
+++ b/lib/ioki/apis/endpoints/endpoints.rb
@@ -17,7 +17,11 @@ module Ioki
       end
     end
 
-    def self.crud_endpoints(resource, model_class:, base_path:, paths: {}, except: nil)
+    # rubocop:disable Metrics/ParameterLists
+    def self.crud_endpoints(resource, model_class:, base_path:, paths: {}, except: nil, only: nil)
+      # rubocop:enable Metrics/ParameterLists
+      raise ArgumentError, 'You cannot set both `only` and `except`. Please only use either one.' if except && only
+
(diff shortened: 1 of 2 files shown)


lib/ioki/apis/endpoints/endpoints.rb
```diff
@@ -34,10 +38,19 @@ def self.crud_endpoints(resource, model_class:, base_path:, paths: {}, except: n
 
         type_key = type.to_s.split('::').last.downcase.to_sym
         name = [Index].include?(type) ? plural : singular
+        action_name = type.to_s.split('::').last.downcase.to_sym
```
rmehner, line 41: What's the difference between the variables `type_key` and `action_name`? They're both assigned with the same expression.

rmehner (approved):

lib/ioki/apis/endpoints/endpoints.rb
phylor: Wow, I really didn't see that :see_no_evil: Fixed!

PR #304, opened 2023-11-13
Rename feature flag

phylor:
`supports_multiple_booking_solutions` is deprecated in favor of `multiple_booking_solutions`.

This might be breaking, but I doubt that anyone is already using it. And we haven't yet released 1.0 😀

### DIFF
--- a/lib/ioki/apis/endpoints/create.rb
+++ b/lib/ioki/apis/endpoints/create.rb
@@ -42,7 +42,7 @@ module Ioki
 
         return if parsed_response.nil?
 
-        model_class.new(parsed_response['data'], response.headers[:etag])
+        model_class.new(parsed_response['data'], response.headers[:etag], show_deprecation_warnings: false)
       end
     end
   end
--- a/lib/ioki/apis/endpoints/delete.rb
+++ b/lib/ioki/apis/endpoints/delete.rb
@@ -33,7 +33,7 @@ module Ioki
 
         return if parsed_response.nil?
 
-        model_class.new(parsed_response['data'])
+        model_class.new(parsed_response['data'], nil, show_deprecation_warnings: false)
       end
     end
   end
--- a/lib/ioki/apis/endpoints/endpoint.rb
+++ b/lib/ioki/apis/endpoints/endpoint.rb
@@ -27,7 +27,7 @@ module Ioki
         )
 
         if model_class
-          model_class.new(parsed_response['data'])
+          model_class.new(parsed_response['data'], nil, show_deprecation_warnings: false)
         else
           parsed_response
         end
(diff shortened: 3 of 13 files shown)


rmehner:
@phylor have you checked with @ans-ioki if this is used somewhere? If yes, we can provide an alias with a deprecation notice to make sure that no one is bitten by this and remove this in a next "release"

This might be the point where we should introduce a proper release :D

I'll approve under the assumption that this attribute isn't used yet

rmehner (approved):

rmehner:
While the filter is nice, I fear it won't lead to action on the user's side. Sure, it's a way for the user to say "We're future proof", but without them knowing what to change, they'll run into issues that aren't very close to the caller side, since the undefined method errors or nil access errors will happen somewhere else where the user uses the serialized object.

So not sure what the purpose of that filter is? Or is this just to silence the warnings after you've fixed the deprecations on users side? If so, how would they notice new deprecations?

I might not be seeing something, so please enlighten me :)

ioki-ruby returns models, not just hashes, why wouldn't it work to overwrite getters/setters there to add a deprecation message once you try to use the methods?

phylor:
> So not sure what the purpose of that filter is? Or is this just to silence the warnings after you've fixed the deprecations on users side? If so, how would they notice new deprecations?

The filter's main purpose is not to serialize deprecated attributes. Let's say you call an API and then dump the result to your database. When serializing the model, you would also get all deprecated attributes. That is clutter I'd like to avoid if possible. I believe that it's still useful to have the filter if you don't object?

> ioki-ruby returns models, not just hashes, why wouldn't it work to overwrite getters/setters there to add a deprecation message once you try to use the methods?

I implemented that in 060ff8245d0097b88b36328239ec9f7ff931e671.

It needs some internal workarounds because otherwise users are spammed with deprecation warnings although they didn't access the deprecated attributes. ioki-ruby is internally also creating models and then also accesses deprecated ones.

rmehner:
> The filter's main purpose is not to serialize deprecated attributes. Let's say you call an API and then dump the result to your database. When serializing the model, you would also get all deprecated attributes. That is clutter I'd like to avoid if possible. I believe that it's still useful to have the filter if you don't object?

Good point. I'd rename the thing to something like `ignore_deprecated_attributes_on_serialization` (naming is hard) to make the intent clear.

phylor:
> Good point. I'd rename the thing to something like `ignore_deprecated_attributes_on_serialization` (naming is hard) to make the intent clear.

We use the filter in `#attributes` and `#serialize`. What about `ignore_deprecated_attributes`? Otherwise it's a bit long and messes with the code layout 😉

rmehner:
> We use the filter in `#attributes` and `#serialize`. What about `ignore_deprecated_attributes`? Otherwise it's a bit long and messes with the code layout 😉

Haha, I feel you. My main point on including "_on_serialization" was that it makes clear that we don't filter / ignore those attributes completely, we just don't serialize them. But honestly this is not a hill I need to die on. If you feel that `ignore_deprecated_attributes` does the thing justice, I'm fine with it :)

PR #308, opened 2023-11-23
Revert "Operator Api | Add matching logs endpoint"

tom-ioki:
Reverts ioki-mobility/ioki-ruby#307


ferdicus (approved):
sleep tight princess 👸🏿

PR #309, opened 2023-11-29
Operator Api | Ride Notification

tom-ioki:
This PR will introduce the ride notification model for the operator api. Also the `rides_notifications` endpoint will be added which can be used to retrieve all notifications that are connected to a specific ride.


pak-ioki (approved):
👍 crisp and clean

PR #313, opened 2023-12-13
Operator Api | Add translated_ride_cancellation_reasons

tom-ioki:
This PR will add the `translated_ride_cancellation_reasons` attribute to the Product model for the operator api.

### DIFF
--- a/lib/ioki/model/operator/multilanguage_string.rb
+++ b/lib/ioki/model/operator/multilanguage_string.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      class MultilanguageString < Base
+        attribute :type,
+                  on:   :read,
+                  type: :string
+
+        attribute :name,
+                  on:   :read,
+                  type: :string
+
+        attribute :translations,
+                  on:         :read,
+                  type:       :array,
+                  class_name: 'Translation'
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/ioki/model/operator/multilanguage_string.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      class MultilanguageString < Base
```
pak-ioki, line 6: MultiLanguageString

tom-ioki: But it's `MultilanguageStringSerializer` in triebwerk 🤔

spec/ioki/model/operator/multilanguage_string_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+RSpec.describe Ioki::Model::Operator::MultilanguageString do
```
pak-ioki, line 3: MultiLanguageString

lib/ioki/model/operator/multilanguage_string.rb
pak-ioki: What does this class do? is it only a string? String as classname is .. well .. too generic

tom-ioki: This is the representation of the `MultilanguageString` model of triebwerk which can be found in the model concerns. For sake of keeping things consistent.

tom-ioki: As discussed let's keep this consistent with triebwerk 👍

pak-ioki (approved):
Lets get this merged!

PR #314, opened 2023-12-15
Operator Api | Add phone calls

tom-ioki:
This PR will add the endpoint `rides_phone_calls` which makes it possible to retrieve all phone calls for a specific ride.


asc-ioki (approved):
Go for it.

PR #321, opened 2024-01-16
Operator Api | Change access for matching_rank

tom-ioki:
This PR will change the access rights for the attribute `matching_rank` of the `TaskList` model.

### DIFF
--- a/lib/ioki/model/operator/task_list.rb
+++ b/lib/ioki/model/operator/task_list.rb
@@ -66,7 +66,7 @@ module Ioki
                   type:           :string
 
         attribute :matching_rank,
-                  on:   :read,
+                  on:   [:create, :read, :update],
                   type: :integer
 
         attribute :paused,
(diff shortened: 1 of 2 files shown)


stw-ioki (approved):
As a simple frontend developer, this looks good to me 👍

stw-ioki (approved):

lib/ioki/model/operator/task_list.rb
```diff
@@ -66,7 +66,7 @@ class TaskList < Base
                   type:           :string
 
         attribute :matching_rank,
-                  on:   :read,
+                  on:   [:create, :read, :update],
```
phylor, line 69: @tom-ioki Should this have a `omit_if_nil_on: [:create, :update]`? (Our specs fail :slightly_smiling_face:).

tom-ioki: hm according to the structure.sql there's no `NOT NULL` constraint for `matching_rank`. Also `null` is a valid type according to the task list create and update schema 🤔 

But the UI does not allow `nil` Pretty confusing.

Let's add the `omit_if_nil_on` part. Sorry for breaking stuff.

PR #322, opened 2024-01-17
Operator Api | Fix Line and Linestop model

tom-ioki:
This PR will fix access for some attributes of the Line and LineStop model.

__Line__
For these attributes `nil` should be omitted on create and update.
- name 
- mode
- skip_time_window_check
- slug

same for
__LineStop__
- tier
- relative_time
- supports_pickup
- supports_dropoff
- supports_pass_through
- station_id

Hope this won't break stuff for you @phylor

### DIFF
--- a/lib/ioki/model/operator/line.rb
+++ b/lib/ioki/model/operator/line.rb
@@ -21,9 +21,8 @@ module Ioki
                   type: :date_time
 
         attribute :name,
-                  on:             [:read, :create, :update],
-                  omit_if_nil_on: [:create, :update],
-                  type:           :string
+                  on:   [:read, :create, :update],
+                  type: :string
 
         attribute :mode,
                   on:             [:read, :create, :update],
--- a/lib/ioki/model/operator/line.rb
+++ b/lib/ioki/model/operator/line.rb
@@ -4,17 +4,54 @@ module Ioki
   module Model
     module Operator
       class Line < Base
-        attribute :type, on: :read, type: :string
-        attribute :id, on: :read, type: :string
-        attribute :created_at, on: :read, type: :date_time
-        attribute :updated_at, on: :read, type: :date_time
-        attribute :name, on: [:read, :create, :update], type: :string
-        attribute :mode, on: [:read, :create, :update], type: :string
-        attribute :route_number, on: [:read, :create, :update], type: :string
-        attribute :skip_time_window_check, on: [:read, :create, :update], type: :boolean
-        attribute :slug, on: [:read, :create, :update], type: :string
-        attribute :variant, on: [:read, :create, :update], type: :string
-        attribute :line_stops, type: :array, on: :read, class_name: 'LineStop'
+        attribute :type,
+                  on:   :read,
+                  type: :string
+
+        attribute :id,
+                  on:   :read,
+                  type: :string
+
+        attribute :created_at,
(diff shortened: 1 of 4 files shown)


lib/ioki/model/operator/line_stop.rb
```diff
@@ -4,19 +4,65 @@ module Ioki
...
+                  type: :date_time
+
+        attribute :tier,
+                  on:             [:read, :create, :update],
+                  omit_if_nil_on: [:create, :update],
+                  type:           :integer
+
+        attribute :relative_time,
+                  on:             [:read, :create, :update],
+                  omit_if_nil_on: [:create, :update],
```
phylor, line 30: `relative_time` is required on `create`.

lib/ioki/model/operator/line_stop.rb
```diff
@@ -4,19 +4,65 @@ module Ioki
...
+                  type:           :boolean
+
+        attribute :supports_pass_through,
+                  on:             [:read, :create, :update],
+                  omit_if_nil_on: [:create, :update],
+                  type:           :boolean
+
+        attribute :station_id,
+                  on:             [:read, :create],
+                  omit_if_nil_on: [:create],
```
phylor, line 60: `station_id` is required.

lib/ioki/model/operator/line_stop.rb
```diff
@@ -4,19 +4,65 @@ module Ioki
...
+                  on:   :read,
+                  type: :date_time
+
+        attribute :updated_at,
+                  on:   :read,
+                  type: :date_time
+
+        attribute :tier,
+                  on:             [:read, :create, :update],
+                  omit_if_nil_on: [:create, :update],
```
phylor, line 25: `tier` is required on `create`.

lib/ioki/model/operator/line.rb
```diff
@@ -4,17 +4,54 @@ module Ioki
...
+                  on:   :read,
+                  type: :date_time
+
+        attribute :updated_at,
+                  on:   :read,
+                  type: :date_time
+
+        attribute :name,
+                  on:             [:read, :create, :update],
+                  omit_if_nil_on: [:create, :update],
```
phylor, line 25: `name` is required on `create` and `update`.

phylor (approved):
Looks fine to me! Also doesn't break anything for us. Thank you for thinking of us! :bow: 

I left comments on some required attributes (according to the schema). Maybe we should always send these even if they are `nil`? The error behavior is probably the same, but it feels better to me personally if we always send in required attributes. Could just be me though, so no veto!

lib/ioki/model/operator/line_stop.rb
tom-ioki: sounds reasonable to me. done 👍

lib/ioki/model/operator/line_stop.rb
tom-ioki: done!

lib/ioki/model/operator/line_stop.rb
tom-ioki: done!

lib/ioki/model/operator/line.rb
tom-ioki: done!

PR #324, opened 2024-01-18
fix: TaskList model

tom-ioki:
This PR will add the `omit_if_nil_on` part to the `matching_rank` attribute of the TaskList model and change access for `matching_rank` for Vehicle and Operator.


phylor (approved):
Perfect, thank you! Also nice boyscouting :tada:

PR #328, opened 2024-01-30
Operator Api | Add endpoints for `ResourceConfiguration`

tom-ioki:
This PR introduces endpoints for the resource `ResourceConfiguration` and the model for this resource.

### DIFF
--- a/lib/ioki/model/operator/vehicle.rb
+++ b/lib/ioki/model/operator/vehicle.rb
@@ -29,6 +29,11 @@ module Ioki
                   on:   :read,
                   type: :string
 
+        attribute :default_resource_configuration,
+                  on:         :read,
+                  type:       :object,
+                  class_name: 'ResourceConfiguration'
+
         attribute :description,
                   on:             [:create, :read, :update],
                   omit_if_nil_on: [:create, :update],
--- a/lib/ioki/model/operator/vehicle.rb
+++ b/lib/ioki/model/operator/vehicle.rb
@@ -103,25 +103,25 @@ module Ioki
                   on:   [:create, :read, :update],
                   type: :string
 
-        attribute :seats,
-                  on:             [:create, :read, :update],
-                  omit_if_nil_on: [:create, :update],
-                  type:           :integer
+        deprecated_attribute :seats,
+                             on:             [:create, :read, :update],
+                             omit_if_nil_on: [:create, :update],
+                             type:           :integer
 
-        attribute :storage_spaces,
-                  on:             [:create, :read, :update],
-                  omit_if_nil_on: [:create, :update],
-                  type:           :integer
+        deprecated_attribute :storage_spaces,
+                             on:             [:create, :read, :update],
+                             omit_if_nil_on: [:create, :update],
+                             type:           :integer
 
-        attribute :walker_bays,
-                  on:             [:create, :read, :update],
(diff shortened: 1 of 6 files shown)


tom-ioki:
@phylor heads up I have added the `deprecated_attribute` flag for the attributes `flag`, `seats`, `walker_bays` and `wheelchair_bays` of the `Vehicle` model. The `default_resource_configuration` is replacing them.

phylor:
> @phylor heads up I have added the `deprecated_attribute` flag for the attributes `flag`, `seats`, `walker_bays` and `wheelchair_bays` of the `Vehicle` model. The `default_resource_configuration` is replacing them.

Thank you for letting me know! :bow:

lib/ioki/model/operator/vehicle.rb
```diff
@@ -29,6 +29,11 @@ class Vehicle < Base
                   on:   :read,
                   type: :string
 
+        attribute :default_resource_configuration,
```
pak-ioki, line 32: why default_resource_configuration? Is there another resource configuration somewhere?

lib/ioki/model/operator/vehicle.rb
```diff
@@ -98,25 +103,25 @@ class Vehicle < Base
...
-        attribute :walker_bays,
-                  on:             [:create, :read, :update],
-                  omit_if_nil_on: [:create, :update],
-                  type:           :integer
-
-        attribute :wheelchair_bays,
-                  on:             [:create, :read, :update],
-                  omit_if_nil_on: [:create, :update],
-                  type:           :integer
+        deprecated_attribute :seats,
```
pak-ioki, line 106: You can also write 
```
attribute :seats
	deprecated: true
        on:             [:create, :read, :update],
	omit_if_nil_on: [:create, :update],
	type:           :integer
```
or can't you?

pak-ioki (commented):
Looks good. I only have 2 small questions

lib/ioki/model/operator/vehicle.rb
tom-ioki: As @phylor mentioned [here](https://github.com/ioki-mobility/ioki-ruby/pull/304#issuecomment-1810483790)
> In the model classes, we define deprecated attributes using the new method deprecated_attribute instead of attribute. This justs sets deprecated: true on the attribute definition, but is more visible than supplying the option to attribute.

It is for better visibility 👍

lib/ioki/model/operator/vehicle.rb
tom-ioki: That's what it is called in the VehicleSerializer. My understanding is, that a vehicle has one default resource configuration. To get all other resource configurations for a vehicle one has to use the new endpoints for resource configuration 👍

ans-ioki (approved):
I see no problems here. @tom-ioki please be aware that right now in Old Stellwerk we are holding back all of this - in old Stellwerk all of this is superadmin only because we have not yet released the vehicle resources as they depend on the also not fully released dynamic passenger options.

tom-ioki:
> I see no problems here. @tom-ioki please be aware that right now in Old Stellwerk we are holding back all of this - in old Stellwerk all of this is superadmin only because we have not yet released the vehicle resources as they depend on the also not fully released dynamic passenger options.

Good to know! Then team operator will wait with the frontend implementation for the resource configuration 👍

PR #329, opened 2024-01-30
Platform API | Add `User#analytics_tracking` and `User#minimum_age_confirmed`

phylor:
Adds two new attributes when creating users via the platform API.


pak-ioki (approved):
LGTM

PR #334, opened 2024-02-08
Operator Api | `RematchingAttempts` and `RematchingSuggestion`

tom-ioki:
This PR will add endpoints for `RematchingAttempts` and `RematchingSuggestion`.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -269,7 +269,7 @@ module Ioki
       ),
       Endpoints.crud_endpoints(
         :rematching_attempt,
-        base_path:   [API_BASE_PATH, 'products', :id],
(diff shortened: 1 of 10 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -255,6 +255,17 @@ class OperatorApi
         base_path:   [API_BASE_PATH, 'products', :id, 'rides', :id],
         path:        'phone_calls',
         model_class: Ioki::Model::Operator::PhoneCall
+      ),
+      Endpoints.crud_endpoints(
+        :rematching_attempt,
+        base_path:   [API_BASE_PATH, 'products', :id],
```
ans-ioki, line 261: I added comments on the other PR - if we have top level access on those objects, we need safe scoping in triebwerk. I assume we will do that there, so i will prematurely mark this PR as "approved".

ans-ioki (approved):

PR #336, opened 2024-02-14
Operator Api | Allow `nil` for `matching_rank`

tom-ioki:
With this PR `nil` will be a valid value for the attribute `matching_rank` of the models `TaskList` and `Vehicle`.

@phylor hope this is fine!


phylor (approved):
:+1:

PR #343, opened 2024-02-28
[Breaking] Change to serialize only dirty attributes

### LINKED ISSUE #342: Extend model layer to handle nil vs unchanges attributes
In some cases, sending `attribute: nil` is different in behavior than not sending the attribute in at all. This is true for both `create` and `update` requests.

In some cases, `nil` might mean that nothing is explicitly set, whereas a default value is used when the attribute is missing in the request.

A client has to be able to indicate which behavior is intended.

Rough first idea: Incorporate ActiveModel::Dirty
... tbd.

phylor:
Closes #342.

This changes `Ioki::Model::Base` to only serialize those attributes, which were explicitly passed to the model.

The current behavior always serializes the whole object:

```ruby
irb(main):001> Ioki::Model::Operator::Line.new(slug: 'my-line').serialize
=>
{:type=>nil,
 :id=>nil,
 :created_at=>nil,
 :updated_at=>nil,
 :name=>nil,
 :mode=>nil,
 :route_number=>nil,
 :skip_time_window_check=>nil,
 :slug=>"my-line",
 :variant=>nil,
 :line_stops=>nil}
```

Using `omit_if_blank_on` or `omit_if_nil_on`, we can avoid serializing certain attributes. We cannot distinguish whether we want to send an attribute with the value `nil`, or not send the attribute at all.

This PR solves that issue. The new behavior only serializes the explicitly passed attributes:

```ruby
irb(main):001> Ioki::Model::Operator::Line.new(slug: 'my-line').serialize
=> {:slug=>"my-line"}
```

We achieve this by including `ActiveModel::Dirty` in `Ioki::Model::Base` and track changes whenever an attribute is set. The serialization then only serializes "dirty" (i.e. changed) attributes.

## Additional Changes

- Add `Ioki::Model::Base.changed_attributes` to return only changed attributes

## Caveats

- This is a breaking change, so we probably should increase our major version.
- If you don't want to send an attribute at all, but you already initialized it, you have to manually clear the dirty state. Example:

  ```ruby
  irb(main):002> line = Ioki::Model::Operator::Line.new(slug: 'my-line')
  => Ioki::Model::Operator::Line:3400 @attributes={:slug=>"my-line"}
  irb(main):003> line.serialize
  => {:slug=>"my-line"}

  # Now I notice that I don't want to send in `slug` at all.
  # I can't set it to `nil`, because then I would still send in `slug: nil`
  irb(main):007> line.slug = nil
  => nil
  irb(main):008> line.serialize
  => {:slug=>nil}

  # The correct way is to use the `ActiveModel::Dirty` API and clear the change state:
  irb(main):009> line.clear_slug_change
  => nil
  irb(main):010> line.serialize
  => {}
  ```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     ioki-ruby (0.14.0)
+      activemodel (>= 6.0.0)
       faraday (>= 1.8, < 3.0)
       faraday_middleware (~> 1.2)
       oauth2 (>= 2.0.2, < 3)
@@ -10,12 +11,27 @@ PATH
 GEM
   remote: https://rubygems.org/
   specs:
+    activemodel (7.1.3.2)
+      activesupport (= 7.1.3.2)
+    activesupport (7.1.3.2)
+      base64
+      bigdecimal
+      concurrent-ruby (~> 1.0, >= 1.0.2)
+      connection_pool (>= 2.2.5)
+      drb
+      i18n (>= 1.6, < 2)
+      minitest (>= 5.1)
+      mutex_m
+      tzinfo (~> 2.0)
     addressable (2.8.6)
       public_suffix (>= 2.0.2, < 6.0)
     ast (2.4.2)
     awesome_print (1.9.2)
+    base64 (0.2.0)
     bigdecimal (3.1.6)
     coderay (1.1.3)
+    concurrent-ruby (1.2.3)
+    connection_pool (2.4.1)
     crack (1.0.0)
       bigdecimal
       rexml
@@ -25,6 +41,8 @@ GEM
       reline (>= 0.3.8)
     diff-lcs (1.5.1)
     docile (1.4.0)
+    drb (2.2.0)
+      ruby2_keywords
     faraday (1.10.3)
       faraday-em_http (~> 1.0)
       faraday-em_synchrony (~> 1.0)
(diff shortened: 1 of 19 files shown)


tom-ioki (approved):
Really nice work! There was no issue running the test suite for operator. Also, updating resources in the app worked pretty smoothly 👌
Looks like `ActiveModel::Dirty` is/was the right choice. Could you run me through the Base class by any chance?

tom-ioki:
Also I noticed, not all attributes of a model come with a `clear_ABC_change` method. Why is that? For example, the vehicle model has no `change_external_id_change` method.

phylor:
> Also I noticed, not all attributes of a model come with a `clear_ABC_change` method. Why is that? For example, the vehicle model has no `change_external_id_change` method.

Works on my machine :see_no_evil: Which exact model did you try?

```ruby
irb(main):003> vehicle = Ioki::Model::Operator::Vehicle.new(external_id: 'test')
=> Ioki::Model::Operator::Vehicle:3440 @attributes={:external_id=>"test"}

irb(main):004> vehicle.serialize
=> {:external_id=>"test"}

irb(main):005> vehicle.clear_external_id_change
=> nil

irb(main):006> vehicle.serialize
=> {}
```

Also quickly checked with the method `attributes` and it looks as every attribute has a `clear_ .. _change` method:

```ruby
irb(main):019> vehicle.methods.select { |m| m.to_s.include?("clear") }.map { |m| m.to_s.sub(/^clear_/, '').sub(/_change$/, '') } - vehicle.attributes.keys.map(&:to_s)
=> ["changes_information", "attribute_changes"]
```

PR #348, opened 2024-03-05
Operator Api | Add endpoint for syncing resource configurations

tom-ioki:
This PR will introduce an endpoint that allows syncing resource configurations between vehicles.


pak-ioki (approved):
Looks good to me

PR #359, opened 2024-04-03
Passenger API | Add "ride series" index

phylor:
This adds an index for "ride series" in the passenger API.

It also changes our pluralization logic. At the moment we deduct the pluralized form of a resource name by appending `s`. For resource names already ending in `s`, that results in a double `s`: `ride_series` -> `ride_seriess`.

The proposed solution is to take `pluralize` from ActiveSupport, which already covers the above case (and some others). For example, `mobility` will result in `mobilities` instead of `mobilitys`.

We do require ActiveSupport now, but we already have ActiveModel, so it's not such a big change.

### DIFF
--- a/lib/ioki/apis/endpoints/create.rb
+++ b/lib/ioki/apis/endpoints/create.rb
@@ -19,7 +19,7 @@ module Ioki
       end
 
       def resource_path_name
-        path || "#{resource}s"
+        path || resource.pluralize
       end
 
       def full_path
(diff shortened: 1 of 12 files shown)


lib/ioki/apis/endpoints/create.rb
```diff
@@ -19,7 +19,7 @@ def name
       end
 
       def resource_path_name
-        path || "#{resource}s"
+        path || resource.pluralize
```
tom-ioki, line 22: Nice 👍

tom-ioki (approved):
Works for me. Thanks for fixing the pluralization issue 🙏

PR #360, opened 2024-04-04
Operator Api | Add driver set password endpoint

tom-ioki:
This PR will introduce the `driver_set_password` endpoint.

### DIFF
--- a/spec/ioki/operator_api_spec.rb
+++ b/spec/ioki/operator_api_spec.rb
@@ -260,19 +260,6 @@ RSpec.describe Ioki::OperatorApi do
     end
   end
 
-  describe '#drivers_set_password(product_id, driver_id, password)' do
-    it 'calls request on the client with expected params' do
-      expect(operator_client).to receive(:request) do |params|
-        expect(params[:url].to_s).to eq('operator/products/0815/drivers/4711/set_password')
-        expect(params[:method]).to eq(:patch)
-        result_with_data
-      end
-
-      expect(operator_client.drivers_set_password('0815', '4711', '123456'))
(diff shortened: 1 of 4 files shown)


spec/ioki/operator_api_spec.rb
```diff
@@ -260,15 +260,16 @@
     end
   end
 
-  describe '#drivers_set_password(product_id, driver_id, password)' do
+  describe '#update_set_password(product_id, driver_id, ...)' do
```
phylor, line 263:
```suggestion
  describe '#update_driver_password(product_id, driver_id, ...)' do
```

Should this have the same name as the method?

phylor (approved):
I was wondering why you are moving `set_password` to a new `Endpoints::Update`. As far as I understand, `Endpoints.custom_endpoints` does not send a body with the requests. So we have to explicitly use `Endpoints::Update` if we want to serialize our model. Is that correct?

If so, we could add a comment to `custom_endpoints` to make that clear.

tom-ioki:
> I was wondering why you are moving `set_password` to a new `Endpoints::Update`. As far as I understand, `Endpoints.custom_endpoints` does not send a body with the requests. So we have to explicitly use `Endpoints::Update` if we want to serialize our model. Is that correct?
> 
> If so, we could add a comment to `custom_endpoints` to make that clear.

Yes, that is correct! I was not able to send in the password with the custom endpoint, hence I introduced the Update endpoint.

Do you mean commenting in the `CustomEndpoints` class, or where should we put that comment?

spec/ioki/operator_api_spec.rb
tom-ioki: yes!

PR #366, opened 2024-04-10
Operator Api | Fix `rematching_suggestions` endpoint

tom-ioki:
This PR will fix the rematching_suggestions endpoint.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -292,9 +292,10 @@ module Ioki
         model_class: Ioki::Model::Operator::RematchingAttempt,
         except:      [:update, :delete]
       ),
-      Endpoints::Index.new(
-        :rematching_suggestions,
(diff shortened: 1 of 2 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -292,9 +292,10 @@ class OperatorApi
         model_class: Ioki::Model::Operator::RematchingAttempt,
         except:      [:update, :delete]
       ),
-      Endpoints::Index.new(
-        :rematching_suggestions,
-        base_path:   [API_BASE_PATH, 'products', :id, 'task_lists', :id],
+      Endpoints.custom_endpoints(
```
phylor, line 295: In case you aren't aware, `Endpoints::ShowSingular.new` could also be used here. It's used for resources which don't have an ID on their own.

No need to change though!

phylor (approved):
Looks good!

I wasn't able to test it as I couldn't find the right product. I trust you tested it 😀

PR #368, opened 2024-04-12
Operator Api | Add `fleet_state` endpoint

tom-ioki:
Make the fleet_state endpoint accessible for the operator api. And add the `task_list_overview` endpoint.
This will replace the `monitoring` and `vehicle_plannings` endpoint 👍 (which will be removed in a followup PR)


phylor (approved):
Looks good! Approval under the assumption that the related PR is going to be merged as well :slightly_smiling_face:

PR #377, opened 2024-05-07
Operator Api | Add `VehicleTelemetry` model

tom-ioki:
This adds the VehicleTelemetry model for the operator api.


phylor (approved):
Wasn't able to test as I couldn't find any test data 🙂 Looks fine though!

PR #379, opened 2024-05-08
Passenger API | Ticketing

phylor:
This adds all model and endpoints related to ticketing to the passenger API.

There are also two additional changes:

- Add support for namespaced models to the passenger API. Currently all model classes of the passenger API are located in `Ioki::Model::Passenger`. For Ticketing it makes sense to namespace all ticketing models in `Ioki::Model::Passenger::Ticketing`. Such subfolders were not yet loaded though (in `lib/ioki/model.rb`).
- Allow `POST` requests with an empty body. You still have to pass `nil` to the method though: `passenger_client.create_my_model('my_id', nil)`

### DIFF
--- a/lib/ioki/apis/endpoints/create.rb
+++ b/lib/ioki/apis/endpoints/create.rb
@@ -29,7 +29,7 @@ module Ioki
       def call(client, model, args = [], options = {})
         outgoing_model_class = @outgoing_model_class || options[:outgoing_model_class] || model_class
 
-        unless model.blank? || model.is_a?(outgoing_model_class)
+        unless @outgoing_model_class == :empty || model.is_a?(outgoing_model_class)
           raise(ArgumentError, "#{model} is not an instance of #{outgoing_model_class}")
         end
 
--- a/lib/ioki/apis/endpoints/create.rb
+++ b/lib/ioki/apis/endpoints/create.rb
@@ -29,14 +29,14 @@ module Ioki
       def call(client, model, args = [], options = {})
         outgoing_model_class = @outgoing_model_class || options[:outgoing_model_class] || model_class
 
-        unless model.is_a?(outgoing_model_class)
+        unless model.blank? || model.is_a?(outgoing_model_class)
           raise(ArgumentError, "#{model} is not an instance of #{outgoing_model_class}")
         end
 
         parsed_response, response = client.request(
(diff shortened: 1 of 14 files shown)


lib/ioki/apis/endpoints/create.rb
```diff
@@ -29,14 +29,14 @@ def full_path
       def call(client, model, args = [], options = {})
         outgoing_model_class = @outgoing_model_class || options[:outgoing_model_class] || model_class
 
-        unless model.is_a?(outgoing_model_class)
+        unless model.blank? || model.is_a?(outgoing_model_class)
```
tom-ioki, line 32: Could this lead to any issues at some point?

lib/ioki/apis/passenger_api.rb
```diff
@@ -230,6 +230,41 @@ class PassengerApi
...
+        path:                 'renewal',
+        model_class:          Ioki::Model::Passenger::Ticketing::Voucher,
+        outgoing_model_class: Ioki::Model::Passenger::Ticketing::VoucherRenewal
+      ),
+      Endpoints::Create.new(
+        :ticketing_voucher_subscription_cancellation,
+        base_path:            [API_BASE_PATH, 'ticketing', 'vouchers', :id],
+        path:                 'subscription_cancellation',
+        model_class:          Ioki::Model::Passenger::Ticketing::Voucher,
+        outgoing_model_class: nil
```
tom-ioki, line 253: ah I see 👍

tom-ioki (approved):
Looks good to me 👍  Just a small question.

lib/ioki/apis/endpoints/create.rb
phylor: Hope not :slightly_smiling_face: Do you have any specific scenario in mind?

There is a behavioral change: currently `passenger_client.create_my_model('my_id', my_model_object)` raises if `my_model_object` is `nil`.

With this PR it would no longer raise, but send an empty body (which most likely would get rejected).

I now changed it so that instead of checking for a `nil` model, it now checks whether the `outgoing_model_class` is `:empty`, which also defines the behavior more clearly.

tom-ioki: Sounds great! And no I had no specific scenario in mind. Just gut feeling 😄

PR #383, opened 2024-05-15
Add `service_starts_at` and `service_ends_at` to TaskList model

tom-ioki:
This PR will add the attributes `service_starts_at` and `service_ends_at` to the TaskList model of the operator api and platform api.


phylor (approved):
🚀

PR #391, opened 2024-06-03
Operator Api | Public Transport Alternatives

tom-ioki:
This adds all models that are necessary for public transport alternatives.

### DIFF
--- a/lib/ioki/model/operator/public_transport_stop.rb
+++ b/lib/ioki/model/operator/public_transport_stop.rb
@@ -53,11 +53,11 @@ module Ioki
                   on:   :read
 
         attribute :location_name,
-                  type: :array,
+                  type: :string,
                   on:   :read
 
         attribute :postal_code,
-                  type: :array,
+                  type: :string,
                   on:   :read
 
         attribute :scheduled_arrival,
--- a/lib/ioki/model/operator/public_transport_stop.rb
+++ b/lib/ioki/model/operator/public_transport_stop.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      class PublicTransportStop < Base
+        attribute :type,
+                  on:   :read,
+                  type: :string
+
+        attribute :city,
+                  type: :string,
+                  on:   :read
+
+        attribute :country,
+                  type: :string,
+                  on:   :read
+
+        attribute :county,
+                  type: :string,
+                  on:   :read
+
+        attribute :current_arrival,
+                  type: :string,
+                  on:   :read
+
+        attribute :current_departure,
+                  type: :string,
+                  on:   :read
+
+        attribute :current_platform,
+                  type: :string,
+                  on:   :read
+
+        attribute :dhid,
+                  type: :string,
+                  on:   :read
+
+        attribute :display_times,
+                  type: :array,
+                  on:   :read
+
+        attribute :hafas_ext_id,
+                  type: :array,
+                  on:   :read
+
+        attribute :lat,
+                  type: :float,
+                  on:   :read
+
+        attribute :lng,
+                  type: :float,
+                  on:   :read
+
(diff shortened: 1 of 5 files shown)


lib/ioki/model/operator/ride.rb
```diff
@@ -28,6 +28,11 @@ class Ride < Base
                   on:   :read,
                   type: :boolean
 
+        attribute :booking_solution,
```
phylor, line 31:
```suggestion
        attribute :booked_solution,
```

Right? 🙈

phylor (commented):
Haven't checked all of it yet, just a quick review so far :slightly_smiling_face:

lib/ioki/model/operator/ride.rb
tom-ioki: yes 😓

lib/ioki/model/operator/public_transport_stop.rb
```diff
@@ -0,0 +1,85 @@
...
+        attribute :country,
+                  type: :string,
+                  on:   :read
+
+        attribute :county,
+                  type: :string,
+                  on:   :read
+
+        attribute :current_arrival,
+                  type: :string,
```
phylor, line 24:
```suggestion
                  type: :date_time,
```

Haven't tested it, but maybe we want to have a `DateTime` object here?

In the passenger API we defined it as `string` as well. Maybe it's more comfortable for you to have a `DateTime` object? Don't really have an opinion here, whatever is easier for you.

lib/ioki/model/operator/public_transport_stop.rb
```diff
@@ -0,0 +1,85 @@
...
+                  on:   :read
+
+        attribute :current_departure,
+                  type: :string,
```
phylor, line 28:
```suggestion
                  type: :date_time,
```

Same as for `current_arrival`

lib/ioki/model/operator/public_transport_stop.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        attribute :current_platform,
+                  type: :string,
+                  on:   :read
+
+        attribute :dhid,
+                  type: :string,
+                  on:   :read
+
+        attribute :display_times,
```
phylor, line 39: These could also be `DateTime` objects if you want them to be.

lib/ioki/model/operator/public_transport_stop.rb
```diff
@@ -0,0 +1,85 @@
...
+        attribute :location_name,
+                  type: :array,
+                  on:   :read
+
+        attribute :postal_code,
+                  type: :array,
+                  on:   :read
+
+        attribute :scheduled_arrival,
+                  type: :string,
```
phylor, line 64:
```suggestion
                  type: :date_time,
```

Depending on your decision above.

lib/ioki/model/operator/public_transport_stop.rb
```diff
@@ -0,0 +1,85 @@
...
+                  on:   :read
+
+        attribute :scheduled_departure,
+                  type: :string,
```
phylor, line 68:
```suggestion
                  type: :date_time,
```

Depending on your decision above.

phylor (approved):
Looks good! I'm not sure if we want the datetime strings to be parsed by ioki-ruby or not. We haven't done that in the passenger API, but maybe it's more comfortable to work with (we don't have to parse them again)?

lib/ioki/model/operator/public_transport_stop.rb
tom-ioki: Oh that's my bad. Of course these should be of type `date_time`. Thanks 🙇

lib/ioki/model/operator/public_transport_stop.rb
tom-ioki: type array is fine 👍

PR #401, opened 2024-07-02
Operator Api | Add endpoints for `Zone`

tom-ioki:
This adds the endpoints for `Zones` and the `Zone` model for the operator api.

### DIFF
--- a/lib/ioki/model/operator/geojson.rb
+++ b/lib/ioki/model/operator/geojson.rb
@@ -6,15 +6,14 @@ module Ioki
       class Geojson < Base
         unvalidated true
 
-        attribute :coordinates, type: :array, on: [:create, :read]
-        attribute :type, type: :string, on: [:create, :read]
+        attribute :coordinates, type: :array, on: [:create, :read, :update]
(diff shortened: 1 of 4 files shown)


phylor (approved):

lib/ioki/model/operator/geojson.rb
```diff
@@ -6,8 +6,8 @@ module Operator
       class Geojson < Base
         unvalidated true
 
-        attribute :coordinates, type: :array, on: :read
-        attribute :type, type: :string, on: :read
+        attribute :coordinates, type: :array, on: [:create, :read]
+        attribute :type, type: :string, on: [:create, :read]
```
phylor, line 10: I don't think you need `type` on `:create` - at least I believe so :slightly_smiling_face:

phylor: Different type here, so we apparently need that. Sorry for the noise!

tom-ioki: no problem 👍

PR #403, opened 2024-07-09
Operator Api | Refine `Area` model

tom-ioki:
This PR will add more attributes to the Area model for the operator api.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -227,6 +227,15 @@ module Ioki
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::Area
       ),
+      Endpoints.custom_endpoints(
+        'areas',
+        actions:     {
+          'drt_area'        => :get,
+          'intermodal_area' => :get
+        },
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -227,6 +227,15 @@ class OperatorApi
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::Area
       ),
+      Endpoints.custom_endpoints(
```
phylor, line 230: This is how we have defined these endpoints in the platform API:

```ruby
      Endpoints::ShowSingular.new(
        :drt_area,
        model_class: Ioki::Model::Platform::Area
      ),
      Endpoints::ShowSingular.new(
        :intermodal_area,
        model_class: Ioki::Model::Platform::Area
      )
```

Which results in the methods:

- `drt_area(product_id)`
- `intermodal_area(product_id)`

In case you don't want to have the `areas_` prefix.

phylor (approved):

lib/ioki/apis/operator_api.rb
tom-ioki: 👍 sounds good. I will change that :)

PR #409, opened 2024-07-16
Operator Api | Add `num_prebooked_rides` attribute to `TaskList` model

tom-ioki:


phylor (approved):
It seems this is only valid for the `overview` endpoint. We're reusing a model here for two different purposes. I think it's fine for now, maybe we need to be vigilant about it in the future.

tom-ioki:
> It seems this is only valid for the `overview` endpoint. We're reusing a model here for two different purposes. I think it's fine for now, maybe we need to be vigilant about it in the future.

Indeed!

PR #416, opened 2024-08-12
Reporting API

phylor:
This adds the basics of the reporting API.

It also improves ioki-ruby to allow parsing different types in an array. Untill now, only one type was supported. Instead of looking only at `class_name` of the attribute definition, we now also check if we have a `type` within the object in the array - if we have, we use that to create a Ruby class.

The following thus creates different classes within the array:

```ruby
[
  { type: 'car' },
  { type: 'bicycle' }
]

# results in
[
  Car.new(..),
  Bicycle.new(..)
]
```

### DIFF
--- a/lib/ioki/model/operator/reporting/report_structure/item_group.rb
+++ b/lib/ioki/model/operator/reporting/report_structure/item_group.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      module Reporting
+        module ReportStructure
+          class ItemGroup < Base
+            attribute :type,
+                      on:   :read,
+                      type: :string
+
+            attribute :name,
+                      on:   :read,
+                      type: :string
+
+            attribute :label,
+                      on:   :read,
+                      type: :string
+
+            attribute :description,
+                      on:   :read,
+                      type: :string
+
+            attribute :items,
+                      on:         :read,
+                      type:       :array,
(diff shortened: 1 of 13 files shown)


lib/ioki/model/operator/reporting/report_structure/item_group.rb
```diff
@@ -0,0 +1,34 @@
...
+                      type: :string
+
+            attribute :description,
+                      on:   :read,
+                      type: :string
+
+            attribute :items,
+                      on:         :read,
+                      type:       :array,
+                      class_name: ['Item', 'ItemGroup', 'Divider']
```
tom-ioki, line 28: Question: I saw there is also a Label class, why is this missing here?

tom-ioki (approved):
Nice one! Just one question.

lib/ioki/model/operator/reporting/report_structure/item_group.rb
phylor: Good catch! :+1: I haven't yet added it because I'm unsure whether we're going to need it. This part of the API is still experimental (and I'm not sure yet how to use the label :slightly_smiling_face:) - so maybe this will disappear again.

I'll keep it out for now - let's see if we need to add it later. Thanks for the find though! :muscle:

tom-ioki: Alright! Thanks for the explanation 👍

PR #441, opened 2024-10-10
Reporting | Add `visible_filter` to reporting column definitions

phylor:
Adds another visibility attribute to column definitions.


tom-ioki (approved):
LGTM

PR #456, opened 2024-11-05
Operator Api | Adjust paths for operator related endpoints

tom-ioki:
This PR will change the paths for the operator related endpoints.


phylor (approved):
:tada:

PR #459, opened 2024-11-22
Operator Api | Endpoints for `PassengerOptions`

tom-ioki:
This adds endpoints for `PassengerOptions` and the `PassengerOption` model


lib/ioki/model/operator/passenger_option.rb
```diff
@@ -0,0 +1,102 @@
...
+                  on:   :read,
+                  type: :date_time
+
+        attribute :slug,
+                  on:   [:read, :create],
+                  type: :string
+
+        attribute :default_value,
+                  on:   [:read, :create, :update],
+                  type: [:string, :boolean, :integer]
```
phylor, line 29: Did you test whether this deserializes integers/booleans correctly? I faintly remember that we had to put `string` at the end, because we start deserializing at the beginning and stop when it works. I believe deserializing as string always works. But I'm not sure if it's implemented as I describe it.

lib/ioki/model/operator/passenger_option.rb
```diff
@@ -0,0 +1,102 @@
...
+
+        attribute :data_type,
+                  on:   [:read, :create],
+                  type: :string
+
+        attribute :option_type,
+                  on:   [:read, :create, :update],
+                  type: :string
+
+        attribute :sort_key,
```
phylor, line 43: I don't find the `sort_key` in the serializer? Maybe I'm looking in the wrong place?

phylor (approved):

lib/ioki/model/operator/passenger_option.rb
tom-ioki: I forgot to add the `sort_key` in triebwerk. But it's fixed now. Thanks!

PR #464, opened 2025-01-14
Operator Api | Add `station` to `LineStop` model

tom-ioki:
This PR introduces `station` to the `LineStop` model.

### DIFF
--- a/lib/ioki/model/operator/line_stop.rb
+++ b/lib/ioki/model/operator/line_stop.rb
@@ -59,6 +59,11 @@ module Ioki
                   on:   [:read, :create],
                   type: :string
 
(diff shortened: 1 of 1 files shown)


tom-ioki:
> Doesn't the API return a `station_id` instead of the whole object?

Yes, it does. I didn't change the api which means there's not really a station object. I added the `station` attribute so that I can assign the station object from the Stellwerk side. Hope I'm making sense right now 😅 . In Stellwerk I request both the index for lines and stations. Then I assign these stations to their corresponding line_stop.

tom-ioki:
Solved it in the operator backend.

PR #468, opened 2025-01-23
Operator Api | Restricted Areas

tom-ioki:
This PR adds endpoints for the `RestrictedArea` resource.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -399,7 +399,8 @@ module Ioki
       Endpoints.crud_endpoints(
         :restricted_area,
         base_path:   [API_BASE_PATH, 'products', :id],
-        model_class: Ioki::Model::Operator::RestrictedArea
+        model_class: Ioki::Model::Operator::RestrictedArea,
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -395,6 +395,11 @@ class OperatorApi
         :blacklisted_travel_combination,
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::BlacklistedTravelCombination
+      ),
+      Endpoints.crud_endpoints(
```
phylor, line 399: If we can't `update` it (I assume?), should we include `except: [:update]`?

phylor (approved):

lib/ioki/apis/operator_api.rb
tom-ioki: Totally missed that one! Thanks for pointing it out. I will exclude it 👍

tom-ioki: Done!

PR #510, opened 2025-03-19
Passenger Api | Add `passenger_note_to_driver` to `Ride` model 

bogn83:
This adds the attribute `passenger_note_to_driver` to the `Ride` model of the passenger API.

A bit suprisingly given the PR title it also adds it to the `ProductFeatures` model in the platform API. So far apparently this is only used for the schalter bootstrap deciding whether or not to show the textarea and in the integration spec in schalter to enable the feature for a test.

### DIFF
--- a/lib/ioki/model/passenger/ride.rb
+++ b/lib/ioki/model/passenger/ride.rb
@@ -47,6 +47,7 @@ module Ioki
         attribute :ticket, on: :read, type: :object, class_name: 'Ticket'
         attribute :offered_solutions, on: :read, type: :array, class_name: 'OfferedSolution'
         attribute :options, on: :create, type: :array, class_name: 'RideOption'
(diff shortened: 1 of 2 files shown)


lib/ioki/model/passenger/ride.rb
```diff
@@ -47,6 +47,7 @@ class Ride < Base
         attribute :ticket, on: :read, type: :object, class_name: 'Ticket'
         attribute :offered_solutions, on: :read, type: :array, class_name: 'OfferedSolution'
         attribute :options, on: :create, type: :array, class_name: 'RideOption'
+        attribute :passenger_note_to_driver, type: :string, on: [:read, :create, :update]
```
bogn83, line 50: Sidenote: schalter is never updating the note for the ride, so `update` in the `on` part is a bit of _you might need it_.

phylor (approved):

PR #511, opened 2025-03-19
Operator Api | add `DriverMulticast` endpoints

tom-ioki:
This adds the `DriverMulticast` model and the endpoints `index`, `show` and `create` for driver multicasts.

### DIFF
--- a/lib/ioki/apis/operator_api.rb
+++ b/lib/ioki/apis/operator_api.rb
@@ -435,6 +435,12 @@ module Ioki
         :venue,
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::Venue
+      ),
+      Endpoints.crud_endpoints(
+        :driver_multicast,
+        base_path:   [API_BASE_PATH, 'products', :id],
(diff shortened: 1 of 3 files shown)


lib/ioki/apis/operator_api.rb
```diff
@@ -401,6 +401,12 @@ class OperatorApi
         base_path:   [API_BASE_PATH, 'products', :id],
         model_class: Ioki::Model::Operator::RestrictedArea,
         except:      [:update]
+      ),
+      Endpoints.crud_endpoints(
+        :driver_multicast,
+        base_path:   [API_BASE_PATH, 'products', :id],
+        model_class: Ioki::Model::Operator::DriverMulticast,
+        except:      [:update, :delete]
```
phylor, line 409: I believe you could also use `only`, if that's a bit easier. Maybe a note for the future, as we're using mostly `except`, but that's more error prone than an allowlist :slightly_smiling_face:

phylor (approved):

lib/ioki/apis/operator_api.rb
tom-ioki: True and good point 👍

PR #535, opened 2025-05-07
Delete operator models

eikes:
These models are not used in the codebase and are not present in the operator API openAPI json schema.

They are mostly the result of a sloppy copy and paste from the platform API models in commit:

https://github.com/ioki-mobility/ioki-ruby/commit/f17abc3e4b2b57db1f43d6c8efa1501708dcb96a

`NotificationSettings` might have been used in the User model but that class now uses an array of the singular `NotificationSetting` class.


phylor (approved):
Looks good to me, but I think @tom-ioki approval is more important here.

tom-ioki (approved):
Yep! Looks good to me 👍 

None of mentioned models are in use.

eikes:
Okay the changes have been merged in another branch now.

PR #562, opened 2025-05-28
Passenger API | Ticketing payment method types

phylor:
Adds a missing accessor for the payment method types used in ticketing to the passenger API.


tom-ioki (approved):
👍

PR #596, opened 2025-07-31
Add seatbelt state to telemetry

tie-ioki:
I noticed that the seatbelt state is missing in the telemetry model. It's present in the Platform API serializer for vehicle telemetry and also an attribute that we want to have in Workflow.

### DIFF
--- a/lib/ioki/model/platform/telemetry.rb
+++ b/lib/ioki/model/platform/telemetry.rb
@@ -152,6 +152,10 @@ module Ioki
                   on:   :read,
                   type: :date_time
 
+        attribute :seatbelt_state,
+                  on:   :read,
+                  type: :string
+
         attribute :speed,
                   on:   :read,
                   type: :float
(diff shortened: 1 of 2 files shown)


lib/ioki/model/platform/telemetry.rb
```diff
@@ -152,6 +152,10 @@ class Telemetry < Base
                   on:   :read,
                   type: :date_time
 
+        attribute :seatbelt_state,
+                  on:   :read,
+                  type: :string
+
```
phylor, line 158:
```suggestion
        attribute :seatbelt_state,
                  on:   :read,
                  type: :string

        attribute :seatbelt_state_source,
                  on:   :read,
                  type: :string

        attribute :seatbelt_state_synced_at,
                  on:   :read,
                  type: :date_time

```

phylor (approved):

lib/ioki/model/platform/telemetry.rb
tie-ioki: Ah right. Thanks, idk how I overlooked that the others all have this

PR #607, opened 2025-08-20
Operator Api | Add more missing features to the `Features` model

tom-ioki:
Unfortunately, I missed two attributes for the features model. This will add the missing attributes.


ferdicus (approved):
my missing babies 👨‍🍼

PR #613, opened 2025-09-01
Operator API | Geocoding

phylor:
Adds endpoints for geocoding, which allows to search for stations and places in the same API request.


tom-ioki (approved):
Looks good to me 👍

PR #622, opened 2025-09-15
Webhook: Add `AdminNotification` model

tom-ioki:


lib/ioki/model/webhooks/admin_notification.rb
```diff
@@ -0,0 +1,18 @@
...
+    module Webhooks
+      class AdminNotification < Base
+        attribute :type, type: :string
+        attribute :id, type: :string
+        attribute :created_at, type: :date_time
+        attribute :updated_at, type: :date_time
+        attribute :admin_notification_type, type: :string
+        attribute :title, type: :string
+        attribute :description, type: :string
+        attribute :read_at, type: :date_time
```
phylor, line 14:
```suggestion
        attribute :read_at, type: :date_time
        attribute :receiver_id, type: :string
```

If you need it?

phylor (approved):
Looks good. Do you also need the `AdminAnnouncement`?

lib/ioki/model/webhooks/admin_notification.rb
tom-ioki: 👍

phylor (approved):

PR #626, opened 2025-09-22
Allow request body for `DeleteSingular`

tom-ioki:
This changes the DeleteSingular endpoint so that it can handle incoming models and allows a request body.

### DIFF
--- a/lib/ioki/client.rb
+++ b/lib/ioki/client.rb
@@ -26,7 +26,12 @@ module Ioki
           options = args.last.is_a?(::Hash) ? args.pop : {}
           model = args.last
 
-          if [Endpoints::Create, Endpoints::Update, Endpoints::UpdateSingular].include?(endpoint.class)
+          if [
+            Endpoints::Create,
+            Endpoints::Update,
+            Endpoints::UpdateSingular,
+            Endpoints::DeleteSingular
+          ].include?(endpoint.class)
             endpoint.call(self, model, args, options)
           elsif endpoint.is_a? Endpoints::Index
             endpoint.call(self, args, options, &block)


lib/ioki/apis/endpoints/delete_singular.rb
```diff
@@ -6,6 +6,27 @@ class DeleteSingular < Delete
       def full_path
         base_path + [resource_path_name]
       end
+
+      def call(client, model, args = [], options = {})
+        outgoing_model_class = @outgoing_model_class || options[:outgoing_model_class] || model_class
+
+        unless model.is_a?(outgoing_model_class)
```
phylor, line 13: `DELETE` requests usually don't have a request body. It does work, but it's not that common. From the [RFC](https://www.rfc-editor.org/rfc/rfc9110.html#name-delete):

> Although request message framing is independent of the method used, content received in a DELETE request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]). A client SHOULD NOT generate content in a DELETE request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.

That being said, we might want to allow `outgoing_model_class` to be `nil`, so that we can do a `DELETE` request to a resource without sending a body, and without getting any response body (`model_class` would then also be `nil`, correct?).

lib/ioki/apis/endpoints/delete_singular.rb
```diff
@@ -6,6 +6,27 @@ class DeleteSingular < Delete
...
+          raise(ArgumentError, "#{model} is not an instance of #{outgoing_model_class}")
+        end
+
+        parsed_response, = client.request(
+          url:    client.build_request_url(*Endpoints.url_elements(full_path, *args)),
+          method: :delete,
+          body:   { data: model.serialize(:delete) },
```
phylor, line 20: I think it would be great if we could still use that endpoint class without a body. We'd need to set `body` to `nil`, if the `model` is not set. What do you think?

phylor (commented):
Looks good! But I think we need a couple of specs testing all the possibilities for this endpoint. For example, a request without body and response, a request with body and response, etc.

PR #708, opened 2026-04-21
Operator | Add `LineState` endpoint

tom-ioki:
This adds the `line_state` endpoint and the `LineState` model.

### DIFF
--- a/lib/ioki/model/operator/line_state.rb
+++ b/lib/ioki/model/operator/line_state.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      class LineState < Base
+        attribute :type,
+                  on:   :read,
+                  type: :string
+
+        attribute :limit_reached,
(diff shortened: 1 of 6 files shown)


lib/ioki/model/operator/line_state.rb
```diff
@@ -0,0 +1,27 @@
...
+                  type: :string
+
+        attribute :limit_reached,
+                  on:   :read,
+                  type: :boolean
+
+        attribute :lines,
+                  on:         :read,
+                  type:       :array,
+                  class_name: 'Line'
```
phylor: I'd personally prefer separate models for `Line` and `LineStop`, because these will probably develop independent of the "root" models. No veto though.

phylor (approved):

lib/ioki/model/operator/line_state.rb
tom-ioki: Yep, totally agree with you. I will do a follow up for line_state and station_state 👍

PR #716, opened 2026-05-07
Operator API | Add recently used stations

phylor:
This adds an endpoint to the operator API to fetch recently used stations of a user.


tom-ioki (approved):
LGTM

PR #722, opened 2026-05-27
Operator API | Aggregations

phylor:
Adds aggregation endpoints to the operator API.

## Breaking change

Currently, when sending a timestamp to the API via a `Ioki::Model`, the timestamp is sent as `to_s`, e.g.: `2026-06-02 11:57:06 +0200`. This format is then rejected by the API as non-ISO8601.

In some projects this was solved by calling `ios8601` on the date object before passing the `Ioki::Model` to the ioki-ruby client. With the following change, this will no longer be necessary.

When sending a `POST`, `PATCH` or `DELETE` request, we now serialize timestamps correctly as ISO8601. To keep the change minimal, we only do that if the attribute is of type `datetime`. We also check whether the value has a `iso8601` method.

Clients which call `iso8601` on their own are not affected. This still works, but they could remove their manual `iso8601` call.

This should therefore not be a breaking change - but still mentioned as potentially breaking.

### DIFF
--- a/lib/ioki/apis/endpoints/create.rb
+++ b/lib/ioki/apis/endpoints/create.rb
@@ -36,7 +36,7 @@ module Ioki
         parsed_response, response = client.request(
           url:    client.build_request_url(*Endpoints.url_elements(full_path, *args)),
           method: :post,
-          body:   { data: model&.serialize(:create) },
+          body:   { data: model&.serialize(:create, format: :json) },
           params: options[:params]
         )
 
--- a/lib/ioki/apis/endpoints/delete.rb
+++ b/lib/ioki/apis/endpoints/delete.rb
@@ -33,7 +33,7 @@ module Ioki
           raise(ArgumentError, "#{model} is not an instance of #{outgoing_model_class}")
         end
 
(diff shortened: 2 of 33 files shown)


tom-ioki (commented):
I tested this in Stellwerk. Unfortunately, I get some errors. It looks like the `Ioki::Model::Operator::Geojson` model is the root cause. The model overrides the `serialize` method (see [here](https://github.com/ioki-mobility/ioki-ruby/blob/bb3eec53014a59f82560d57f15bd6d7ec3f1438f/lib/ioki/model/operator/geojson.rb#L14-L19)) I think this method also has to reflect the new `format` handling?

phylor:
> It looks like the `Ioki::Model::Operator::Geojson` model is the root cause. The model overrides the `serialize` method (see [here](https://github.com/ioki-mobility/ioki-ruby/blob/bb3eec53014a59f82560d57f15bd6d7ec3f1438f/lib/ioki/model/operator/geojson.rb#L14-L19)) I think this method also has to reflect the new `format` handling?

Yes, missed that method. Fixed!

tom-ioki (approved):
Nice! Thanks for taking care of this 👍 

Tests are all green now!

PR #723, opened 2026-05-27
Switch mailer to use new recipient options in api

tie-ioki:
As part of https://github.com/dbdrive/sicherungskasten/issues/1157 we will need to send emails to admins. To facilitate this, we changed the API to receive a more generic `recipient_id` and `recipient_type` over a `user_id` here (https://github.com/dbdrive/triebwerk/pull/20309). This brings ioki ruby back in line with that.


phylor (approved):
Nice!

PR #727, opened 2026-06-08
Operator API | Aggregation rename and new `/totals` endpoint

phylor:
- Renames the `/aggregate` endpoint to `/series`
- Adds new endpoint `/totals`

Although this is a breaking change, the only current consumer has adapted to the changes already.


tom-ioki (approved):
👍

PR #730, opened 2026-06-09
Fix receiver/recipient mixup

tie-ioki:
I mixed up the naming because the fields are `receiver_id` and `receiver_type`, but the constraint is called `required_recipient_id`...


phylor (approved):
A reviewer could have noticed, I guess 🙂🤷

PR #733, opened 2026-06-12
Add eligibility group memberships to operator api

tie-ioki:
- Pays into https://github.com/dbdrive/triebwerk/issues/20300
- Depends on https://github.com/dbdrive/triebwerk/pull/20470 (Do not merge label until this is merged)

### DIFF
--- a/lib/ioki/model/operator/eligibility_group_membership.rb
+++ b/lib/ioki/model/operator/eligibility_group_membership.rb
@@ -4,14 +4,6 @@ module Ioki
   module Model
     module Operator
       class EligibilityGroupMembership < Base
-        attribute :email_address,
-                  on:   :create,
-                  type: :string
-
-        attribute :user_segment_slug,
-                  on:   :create,
-                  type: :string
-
         attribute :id,
                   on:   :read,
                   type: :string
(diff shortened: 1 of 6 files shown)


lib/ioki/model/operator/eligibility_group_membership.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Ioki
+  module Model
+    module Operator
+      class EligibilityGroupMembership < Base
+        attribute :email_address,
```
phylor: This model is a bit convoluted. It combines the schema used in the `POST` request with the response serializer. We usually only do that if both are named identically and match conceptually (e.g. `POST` `User` schema returns a `User`). In this context, it seems the two models are conceptually different.

Personally I'd split them up:

1. `Ioki::Model::Operator::EligibilityGroupMembershipManuallyAssignByEmailAddress`
2. `Ioki::Model::Operator::EligibilityGroupMembership`

You should be able to set the schema as `outgoing_model_class` in the endpoint configuration.

phylor (approved):

phylor (approved):
Thank you! 🎉

phylor (approved):

PR #745, opened 2026-07-05
Operator API | Change No Show Blocks to Service Violation Blocks

anm-ioki:
The Operator API now has new endpoints for service violation blocks to replace the existing no show blocks. These changes use the new endpoints and rename the models accordingly.


tom-ioki (approved):
This looks fine. However this is a breaking change. So we need to be cautious merging this 👍

phylor (approved):

PR #746, opened 2026-07-06
Fix exception caused when marshalling models

phylor:
Fixes exception: `no _dump_data is defined for class Proc`

Occurs when `Ioki` models are involved in caching, and therefore marshalled. Until now we created the deprecator on every model, but it should in fact live on the gem module instead.

For non-Rails projects, we use `Kernel` as a deprecator. We only call `warn` on the deprecator, which also exists on `Kernel`.

### DIFF
--- a/lib/ioki.rb
+++ b/lib/ioki.rb
@@ -34,5 +34,13 @@ module Ioki
     def configure
       yield(config)
     end
+
+    def deprecator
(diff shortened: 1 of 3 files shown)


tom-ioki (approved):
Sounds like a reasonable change 👍 Thank you!

PR #753, opened 2026-07-20
Operator API | Change No Shows to Service Violations

anm-ioki:
Similar to https://github.com/ioki-mobility/ioki-ruby/pull/745

The Operator API now has new endpoints for service violations to replace the existing no shows. These changes use the new endpoints and rename the models accordingly. Additionally, there's a new response attribute for the type of service violation, which for now will just be 'no_show' until new functionality is added to triebwerk.


tog-ioki (approved):
Looks good. I assume, team operator knows how to handle this change. :)

PR #770, opened 2026-09-15
Payment service providers

phylor:
Adds models for payment service providers to platform, operator and passenger API.

Also adds support for nested models in the platform API.


tom-ioki (approved):

phylor:
@tom-ioki Please have another look into 54f00a4262429f6c23413e79876e1c8bd1eaebc5. Had to change the definition and fix the parsing of the class names. Sorry!

The problem was that for class names in an array, we only considered the last part of the `type`, e.g. `logpay` when the type is `payment_service_provider/logpay`. Now we also support parsing the whole type with all namespaces.

tom-ioki:
> @tom-ioki Please have another look into [54f00a4](https://github.com/ioki-mobility/ioki-ruby/commit/54f00a4262429f6c23413e79876e1c8bd1eaebc5). Had to change the definition and fix the parsing of the class names. Sorry!
> 
> The problem was that for class names in an array, we only considered the last part of the `type`, e.g. `logpay` when the type is `payment_service_provider/logpay`. Now we also support parsing the whole type with all namespaces.

Ah I see! Alright 👍

